PR #28, opened 2015-06-02
Networking and Production Checklist guides

michaelklishin:
This introduces two new documentation guides:

 * Everything related to networking
 * Production check list

The goal is not to make the guides as extensive as possible but to offer a starting point. We will expand both guides over time based on user feedback and what support questions we get most often.

### DIFF
--- a/site/networking.xml
+++ b/site/networking.xml
@@ -579,6 +579,17 @@ sysctl -w fs.file-max 200000
       </p>
     </doc:section>
 
+    <doc:section name="nettick">
+      <doc:heading>Net Tick Time</doc:heading>
+
+      <p>
+        <a href="/heartbeats.html">Heartbeats</a> are used to detect peer or connection failure
+        between clients and RabbitMQ nodes. <a href="/net_ticktime.html">net_ticktime</a> serves
+        the same purpose but for cluster node communication. Values lower than 5 (seconds) are
+        may result in false positive and are not recommended.
+      </p>
+    </doc:section>
+
     <doc:section name="dns">
       <doc:heading>Hostname Resolution and DNS</doc:heading>
 
@@ -589,7 +600,7 @@ sysctl -w fs.file-max 200000
         covers most common issues associated with that.
       </p>
 
-      
+
       <doc:subsection name="dns-resolution-by-clients">
         <doc:heading>Performed by Client Libraries</doc:heading>
 
@@ -624,7 +635,7 @@ sysctl -w fs.file-max 200000
 
       <doc:subsection name="dns-reverse-dns-lookups">
         <doc:heading>Reverse DNS Lookups</doc:heading>
-      
+
         <p>
           If the <code>rabbit.reverse_dns_lookups</code> configuration option is set to <code>true</code>,
           RabbitMQ will perform reverse DNS lookups for client IP addresses and list hostnames
--- a/site/networking.xml
+++ b/site/networking.xml
@@ -457,7 +457,7 @@ RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+A 128"
             <dt><code>net.ipv4.tcp_max_syn_backlog</code></dt>
             <dd>
               Maximum number of remembered connection requests which
-              still not receive an acknowledgment yet from
+              did not receive an acknowledgment yet from
               connecting client. Default is 128, max value is 65535. 4096 and 8192 are
               recommended starting values when optimising for throughput.
             </dd>
--- a/site/networking.xml
+++ b/site/networking.xml
@@ -380,8 +380,9 @@ RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+A 128"
           but is also small enough to not make much difference. When the number reaches tens of thousands
           or more, it is important to make sure that the server can accept inbound connections.
           Unaccepted TCP connections are put into a queue with bounded length. This length has to be
-          sufficient to account for peak load hours and possible spikes (e. g. mass client
-          disconnects due to a network interruption). This is configured using the <code>rabbit.tcp_listen_options.backlog</code>
+          sufficient to account for peak load hours and possible spikes, for instance, when many clients
+          disconnect due to a network interruption or choose to reconnect.
+          This is configured using the <code>rabbit.tcp_listen_options.backlog</code>
           option:
 
 <pre class="code">
--- a/site/networking.xml
+++ b/site/networking.xml
@@ -74,7 +74,7 @@ limitations under the License.
           For RabbitMQ to accept client connections, it needs to bind to one or more
           interfaces and listen on (protocol-specific) ports. The interfaces are
           configured using the <code>rabbit.tcp_listeners</code> config option.
-          By default, RabbitMQ will listen on all available interfaces.
+          By default, RabbitMQ will listen on port 5672 on all available interfaces.
         </p>
 
         <p>
--- a/site/networking.xml
+++ b/site/networking.xml
@@ -482,7 +482,7 @@ RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+A 128"
           </dl>
 
 
-          Note that default values for these vary between Linux kernel releases.
+          Note that default values for these vary between Linux kernel releases and distributions.
           Using a recent kernel (3.9 or later) is recommended.
         </p>
       </doc:subsection>
--- a/site/networking.xml
+++ b/site/networking.xml
@@ -468,7 +468,7 @@ RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+A 128"
               and <code>net.ipv4.tcp_keepalive_probes</code> configure TCP keepalive.
 
               AMQP 0-9-1 and STOMP have <a href="/heartbeats.html">Heartbeats</a> which partially
-              undo its effect, namely that it can take minutes to detect a unresponsive peer,
+              undo its effect, namely that it can take minutes to detect an unresponsive peer,
               e.g. in case of a hardware or power failure.
 
               When enabling TCP keepalive, we recommend setting heartbeat timeout to 8-20 seconds.
--- a/site/networking.xml
+++ b/site/networking.xml
@@ -590,6 +590,30 @@ sysctl -w fs.file-max 200000
       </p>
     </doc:section>
 
+    <doc:section name="nettick">
+      <doc:heading>Connection Handshake Timeout</doc:heading>
+
+      <p>
+        RabbitMQ has a timeout for connection handshake, 10 seconds by
+        default. When clients run in heavy constrainted environments,
+        it may be necessary to increase the timeout. This can be done via
+        the <code>rabbit.handshake_timeout</code> (in milliseconds):
+
+<pre class="code">
+[
+  {rabbit, [
+    %% 20 seconds
+    {handshake_timeout, 20000}
+  ]}
+].
+</pre>
+
+        It should be pointed out that this is only necessary with very constrained
+        clients and networks. Handshake timeouts in other circumstances indicate
+        a problem elsewhere.
+      </p>
+    </doc:section>
+
     <doc:section name="dns">
       <doc:heading>Hostname Resolution and DNS</doc:heading>
 
--- a/site/networking.xml
+++ b/site/networking.xml
@@ -586,7 +586,7 @@ sysctl -w fs.file-max 200000
       <p>
         <a href="/heartbeats.html">Heartbeats</a> are used to detect peer or connection failure
         between clients and RabbitMQ nodes. <a href="/net_ticktime.html">net_ticktime</a> serves
-        the same purpose but for cluster node communication. Values lower than 5 (seconds) are
+        the same purpose but for cluster node communication. Values lower than 5 (seconds)
         may result in false positive and are not recommended.
       </p>
     </doc:section>
--- a/site/networking.xml
+++ b/site/networking.xml
@@ -268,7 +268,7 @@ RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+A 128"
           When optimising for the number of concurrent connections,
           making sure your system has enough file descriptors to
           support not only client connections but also files the node
-          may use. To calculate a ball park limit, multiply the number
+          may use. To calculate a ballpark limit, multiply the number
           of connections per node by 1.5. For example, to support 100,000
           connections, set the limit to 150,000.
 
--- a/site/networking.xml
+++ b/site/networking.xml
@@ -409,7 +409,7 @@ RABBITMQ_SERVER_ADDITIONAL_ERL_ARGS="+A 128"
           Some are directly related to networking (e.g. TCP settings), others
           affect TCP sockets as well as other things (e.g. open file handles limit).
 
-          Understanding these limits is important, as they may changing depending on
+          Understanding these limits is important, as they may change depending on
           the workload.
         </p>
       </doc:subsection>
--- a/site/networking.xml
+++ b/site/networking.xml
@@ -565,14 +565,6 @@ sysctl -w fs.file-max 200000
             although using <a href="/heartbeats.html">heartbeats</a> is still recommended over
             this option.
           </dd>
-          <dt><code>rabbit.tcp_listen_options.reuseaddr</code></dt>
-          <dd>
-            Sets <code>SO_REUSEADDR</code> on the sockets, which allows a port belonging
-            to a connection in the <code>TIME_WAIT</code> state to be reused for other
-            connections. Enabled by default, highly recommended for environments with
-            large numbers of concurrent connections and/or connection churn (short-lived
-            clients).
-          </dd>
         </dl>
 
       </doc:subsection>
--- a/site/networking.xml
+++ b/site/networking.xml
@@ -1,0 +1,186 @@
+<?xml-stylesheet type="text/xml" href="page.xsl"?>
+<!--
+Copyright (C) 2007-2015 Pivotal Software, Inc.
+
+All rights reserved. This program and the accompanying materials
+are made available under the terms of the under the Apache License,
+Version 2.0 (the "License”); you may not use this file except in compliance
+with the License. You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<html xmlns="http://www.w3.org/1999/xhtml"
+      xmlns:doc="http://www.rabbitmq.com/namespaces/ad-hoc/doc"
+      xmlns:xi="http://www.w3.org/2003/XInclude"
+      xmlns:x="http://www.rabbitmq.com/2011/extensions">
+  <head>
+    <title>Networking and RabbitMQ</title>
+  </head>
+  <body>
+    <doc:section name="intro">
+      <doc:heading>Introduction</doc:heading>
+
+      <doc:subsection name="intro">
+        <p>
+          Clients communicate with RabbitMQ over the network. All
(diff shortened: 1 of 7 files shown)


michaelklishin:
@dumbbell this is ready for review. Note that the goal is not to make the guides as extensive as possible but to offer a starting point. We will expand both guides over time based on user feedback and what support questions we get most often.

Branched off of `live`.

site/production-checklist.xml
```diff
@@ -0,0 +1,290 @@
...
+        </p>
+        <p>
+          It is recommended to use a separate user per application. For example, if you
+          have a mobile app, a Web app, and a data aggregation system, you'd have 3
+          separate users. This makes a number of things easier:
+
+          <ul>
+            <li>Correlating client connections with applications</li>
+            <li>Using <a href="/access-control.html">fine-grained permissions</a></li>
+            <li>Credentials roll-over in case of a breach or periodically</li>
```
videlalvaro: incomplete sentence

site/production-checklist.xml
```diff
@@ -0,0 +1,290 @@
...
+          can be set to <code>{mem_relative, N}</code> to make it calculated
+          as a percentage of the RAM limit. For example, use <code>{mem_relative, 0.5}</code>
+          for 50%, <code>{mem_relative, 0.25}</code> for 25%, and so on.
+        </p>
+      </doc:subsection>
+
+      <doc:subsection name="resource-limits-file-handle-limit">
+        <doc:heading>Open File Handles Limit</doc:heading>
+        <p>
+          Oprating systems limit maximum number of concurrently open file handles, which
```
videlalvaro: "oprating" typo

site/production-checklist.xml
```diff
@@ -0,0 +1,290 @@
...
+      </doc:subsection>
+
+      <doc:subsection name="security-considerations-tls">
+        <doc:heading>TLS</doc:heading>
+        <p>
+          We recommend using <a href="/ssl.html">TLS connections</a> when possible,
+          at least to encrypt traffic. Peer verification (authentication) is also recommended.
+          Development and QA environments can use <a href="https://github.com/michaelklishin/tls-gen/">self-signed TLS certificates</a>.
+          Self-signed certificates can be appropriate in production environments when
+          RabbitMQ and all applications run on a trusted network or iso.
```
videlalvaro: iso as in "isolated" or?

site/production-checklist.xml
```diff
@@ -0,0 +1,290 @@
...
+          <ul>
+            <li>Expected throughput</li>
+            <li>Expected replication (number of mirrors)</li>
+            <li>Data locality</li>
+          </ul>
+        </p>
+        <p>
+          Since clients can connect to any node, RabbitMQ may need to perform inter-cluster
+          routing of messages and internal operations. Try making consumers and producers
+          connect to the same node, if possible: this will reduce inter-node traffic.
```
videlalvaro: If I do this, then why would I use a cluster besides HA? What about load balancing?

site/production-checklist.xml
```diff
@@ -0,0 +1,290 @@
...
+          Equally helpful would be making consumers connect to the node that currently hosts
+          queue master (can be inferred using <a href="/management.html">HTTP API</a>).
+          When data locality is taken into consideration, it
+          <a href="http://blog.pivotal.io/pivotal/products/rabbitmq-hits-one-million-messages-per-second-on-google-compute-engine"></a>
```
videlalvaro: missing text inside the `<a />`

site/production-checklist.xml
michaelklishin: It is complete. Credentials may need rolling over in case of a breach or periodically.

site/networking.xml
```diff
@@ -0,0 +1,671 @@
...
+
+    <doc:section name="interfaces">
+      <doc:heading>Network Interfaces</doc:heading>
+
+      <doc:subsection name="multiple-interfaces">
+        <p>
+          For RabbitMQ to accept client connections, it needs to bind to one or more
+          interfaces and listen on (protocol-specific) ports. The interfaces are
+          configured using the <code>rabbit.tcp_listeners</code> config option.
+          By default, RabbitMQ will listen on all available interfaces.
```
videlalvaro: Perhaps change to "By default, RabbitMQ will listen on port 5672 on all available interfaces."

site/production-checklist.xml
michaelklishin: If you don't need mirroring and want to provision N separate nodes individually, you probably don't.  This is not a guide about why you may want to use clustering.

site/production-checklist.xml
videlalvaro: Then perhaps change the order, since "in case of a breach" should be the less common occurring event. A matter of preference anyway…

site/networking.xml
```diff
@@ -0,0 +1,671 @@
...
+        <doc:heading>Open File Handles Limit</doc:heading>
+        <p>
+          Most operating systems limit the number of file handles that
+          can be opened at the same time. <a href="http://docs.basho.com/riak/latest/ops/tuning/linux/">How
+          the limit is configured</a> varies from OS to OS.
+
+          When optimising for the number of concurrent connections,
+          making sure your system has enough file descriptors to
+          support not only client connections but also files the node
+          may use. To calculate a ball park limit, multiply the number
```
videlalvaro: ballpark is just one word

site/networking.xml
```diff
@@ -0,0 +1,671 @@
...
+          <dl>
+            <dt><code>fs.file-max</code></dt>
+            <dd>
+              Max number of files the kernel will allocate. Limits and current value
+              can be inspected using <code>/proc/sys/fs/file-nr</code>.
+            </dd>
+
+            <dt><code>net.ipv4.ip_local_port_range</code></dt>
+            <dd>
+              Local IP port range, define as a pair of values. The range must .
```
videlalvaro: must?

site/networking.xml
```diff
@@ -0,0 +1,671 @@
...
+      </p>
+    </doc:section>
+
+    <doc:section name="nettick">
+      <doc:heading>Net Tick Time</doc:heading>
+
+      <p>
+        <a href="/heartbeats.html">Heartbeats</a> are used to detect peer or connection failure
+        between clients and RabbitMQ nodes. <a href="/net_ticktime.html">net_ticktime</a> serves
+        the same purpose but for cluster node communication. Values lower than 5 (seconds) are
```
videlalvaro: I think "are" is not needed here

site/networking.xml
```diff
@@ -0,0 +1,672 @@
...
+  ]}
+].
+</pre>
+        </p>
+      </doc:subsection>
+
+      <doc:subsection name="single-stack-ipv4">
+        <doc:heading>Listening on IPv4 Interfaces Only</doc:heading>
+        <p>
+          In this example RabbitMQ will listen on an IPv4 interface only:
```
dumbbell: As the guide primarily targets Linux, I would add a paragraph about the fact that this configuration on Linux (at least Debian, but probably other popular distributions) will listen to both IPv4 and IPv6. The `"127.0.0.1"` entry above is ignored by RabbitMQ on Linux because `"::1"` already takes care of it.

See: https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_networking.erl#L515

site/networking.xml
```diff
@@ -0,0 +1,696 @@
...
+          RAM idle machine uses but this is a reasonable trade-off.
+        </p>
+      </doc:subsection>
+
+      <doc:subsection name="tuning-for-large-number-of-connections-tcp-buffer-size">
+        <doc:heading>TCP Buffer Size</doc:heading>
+        <p>
+          See the section above for an overview.
+          it is possible to decrease buffer size using the <code>rabbit.tcp_listen_options</code> to reduce
+          the amount of RAM by the server used per connection. This is typically necessary
```
dumbbell: The final dot `.` is missing at the end of the last sentence.

michaelklishin:
One thing that's missing: the pages are not linked to from any other pages. Will do that later today.

site/networking.xml
```diff
@@ -0,0 +1,696 @@
...
+        <doc:heading>TCP Buffer Size</doc:heading>
+        <p>
+          This is one of the key tunable parameters. Every TCP connection has buffers
+          allocated for it. Generally speaking, the larger these buffers are, the more RAM
+          is used per connection and better the throughput. On Linux, the OS will automatically
+          tune TCP buffer size by default, typically settling on a value between 80 and 120 KB.
+          For maximum throughput, it is possible to increase buffer size using the <code>rabbit.tcp_listen_options</code>.
+        </p>
+        <p>
+          The following example sets TCP buffers to 192K:
```
dumbbell: The unit is missing. It should be "192 KiB" if we follow the SI units. I'm fine with "192 KB" otherwise (with a space between the number and the unit).

site/networking.xml
```diff
@@ -0,0 +1,696 @@
...
+
+      <doc:subsection name="tuning-for-large-number-of-connections-tcp-buffer-size">
+        <doc:heading>TCP Buffer Size</doc:heading>
+        <p>
+          See the section above for an overview.
+          it is possible to decrease buffer size using the <code>rabbit.tcp_listen_options</code> to reduce
+          the amount of RAM by the server used per connection. This is typically necessary
+        </p>
+        <p>
+          The following example sets TCP buffers to 32K:
```
dumbbell: Likewise, the unit is missing.

site/networking.xml
```diff
@@ -0,0 +1,696 @@
...
+  ]}
+].
+</pre>
+
+         Note that lower TCP buffer sizes will result in a significant throughput drop,
+         so an optional value between throughput and per-connection RAM use needs to be
+         found for every workload.
+
+         Setting send and receive buffer sizes to different values is dangerous
+         and is not recommended. Values lower than 8K are not recommended.
```
dumbbell: Likewise, the unit is missing.

site/networking.xml
```diff
@@ -0,0 +1,696 @@
...
+
+      <doc:subsection name="tuning-for-large-number-of-connections-connection-backlog">
+        <doc:heading>Connection Backlog</doc:heading>
+        <p>
+          With a low number of clients, new connection rate is very unevenly distributed
+          but is also small enough to not make much difference. When the number reaches tens of thousands
+          or more, it is important to make sure that the server can accept inbound connections.
+          Unaccepted TCP connections are put into a queue with bounded length. This length has to be
+          sufficient to account for peak load hours and possible spikes (e. g. mass client
+          disconnects due to a network interruption). This is configured using the <code>rabbit.tcp_listen_options.backlog</code>
```
dumbbell: I would write "massive client reconnections" here.

site/networking.xml
```diff
@@ -0,0 +1,697 @@
...
+[
+  {rabbit, [
+    {tcp_listen_options, [{backlog, 4096}]}
+  ]}
+].
+</pre>
+
+           Default value is 128. When pending connection queue length grows beyond this value,
+           connections will be rejected by the operating system. See also <code>net.core.somaxconn</code>
+           in the kernel tuning section.
```
dumbbell: You could use the same name as the section, "OS Level Tuning", and put a link to it, even if it's just below.

site/networking.xml
```diff
@@ -0,0 +1,697 @@
...
+      <doc:heading>OS Level Tuning</doc:heading>
+
+      <doc:subsection name="os-tuning-intro">
+        <p>
+          Operating system settings can affect operation of RabbitMQ.
+          Some are directly related to networking (e.g. TCP settings), others
+          affect TCP sockets as well as other things (e.g. open file handles limit).
+
+          Understanding these limits is important, as they may changing depending on
+          the workload.
```
dumbbell: There is a mistake in the sentence "they may changing depending the workload". Did you mean "they may change, depending (...)"?

site/networking.xml
```diff
@@ -0,0 +1,698 @@
...
+              will stay in the TIME_WAIT state. Recommended for cases when a large
+              number of concurrent connections is expected.
+            </dd>
+
+            <dt><code>net.core.somaxconn</code></dt>
+            <dd>
+              Size of the listen queue (how many connections are in
+              the process of being established at the same time).
+              Default is 128. Increase to 4096 or higher to support
+              inbound connection bursts, e.g. when clients re-connect
```
dumbbell: I think you can drop the `-` in "reconnect".

site/networking.xml
```diff
@@ -0,0 +1,698 @@
...
+              en masse.
+            </dd>
+
+            <dt><code>net.ipv4.tcp_max_syn_backlog</code></dt>
+            <dd>
+              Maximum number of remembered connection requests, which
+              are still did not receive an acknowledgment from
```
dumbbell: There is a problem with "which are still did not receive (...)".

site/networking.xml
```diff
@@ -0,0 +1,698 @@
...
+              recommended starting values when optimising for throughput.
+            </dd>
+
+            <dt><code>net.ipv4.tcp_keepalive_*</code></dt>
+            <dd>
+              <code>net.ipv4.tcp_keepalive_time</code>, <code>net.ipv4.tcp_keepalive_intvl</code>,
+              and <code>net.ipv4.tcp_keepalive_probes</code> configure TCP keepalive.
+
+              AMQP 0-9-1 and STOMP have <a href="/heartbeats.html">Heartbeats</a> which partially
+              undo its effect, namely that it can take minutes to detect a unresponsive peer,
```
dumbbell: "**an** unresponsive"?

site/networking.xml
```diff
@@ -0,0 +1,698 @@
...
+
+            <dt><code>net.ipv4.conf.default.rp_filter</code></dt>
+            <dd>
+              Enabled reverse path filtering. If <a href="http://en.wikipedia.org/wiki/IP_address_spoofing">IP address spoofing</a>
+              is not a concern for your system, disable it.
+            </dd>
+          </dl>
+
+
+          Note that default values for these vary between Linux kernel releases.
```
dumbbell: ... and, I think, Linux distributions.

site/networking.xml
```diff
@@ -0,0 +1,698 @@
...
+              the process of being established at the same time).
+              Default is 128. Increase to 4096 or higher to support
+              inbound connection bursts, e.g. when clients reconnect
+              en masse.
+            </dd>
+
+            <dt><code>net.ipv4.tcp_max_syn_backlog</code></dt>
+            <dd>
+              Maximum number of remembered connection requests which
+              still did not receive an acknowledgment from
```
dumbbell: "which did not receive an acknowledgment yet"

site/networking.xml
```diff
@@ -0,0 +1,698 @@
...
+          <dt><code>rabbit.tcp_listen_options.nodelay</code></dt>
+          <dd>
+            When enabled, disables
+            <a href="http://en.wikipedia.org/wiki/Nagle's_algorithm">Nagle's algorithm</a>.
+            Default is true. Highly recommended for most users.
+          </dd>
+          <dt><code>rabbit.tcp_listen_options.sndbuf</code></dt>
+          <dd>
+            See TCP buffers discussion earlier in this guide. Default value is
+            automatically tuned by the OS, typically in the 88K to 128K range on
```
dumbbell: MIssign units.

site/networking.xml
```diff
@@ -0,0 +1,698 @@
...
+            When set to <code>{true, N}</code>, sets
+            the timeout in seconds for flushing unsent data when the (server) socket is
+            closed.
+          </dd>
+          <dt><code>rabbit.tcp_listen_options.keepalive</code></dt>
+          <dd>
+            When set to <code>true</code>, enables TCP keepalives (see above).
+            Default is <code>false</code>. Makes sense for environments where
+            connections can go idle for a long time (at least 10 minutes),
+            although <a href="/heartbeats.html">heartbeats</a> is still recommended over
```
dumbbell: "are still recommended"

site/networking.xml
```diff
@@ -0,0 +1,698 @@
...
+            this option.
+          </dd>
+          <dt><code>rabbit.tcp_listen_options.reuseaddr</code></dt>
+          <dd>
+            Sets <code>SO_REUSEADDR</code> on the sockets, which allows a port belonging
+            to a connection in the <code>TIME_WAIT</code> state to be reused for other
+            connections. Enabled by default, highly recommended for environments with
+            large numbers of concurrent connections and/or connection churn (short-lived
+            clients).
```
dumbbell: This option only affects the listening socket (ie. what should we do when RabbitMQ is restarted: if this is `false`, RabbitMQ will probably fail to bind the TCP port without waiting for several seconds), not the accepted connections.

I believe we do not need to document this variable in this guide.

site/networking.xml
```diff
@@ -0,0 +1,698 @@
...
+              the process of being established at the same time).
+              Default is 128. Increase to 4096 or higher to support
+              inbound connection bursts, e.g. when clients reconnect
+              en masse.
+            </dd>
+
+            <dt><code>net.ipv4.tcp_max_syn_backlog</code></dt>
+            <dd>
+              Maximum number of remembered connection requests which
+              still not receive an acknowledgment yet from
```
dumbbell: s/still/did/ :-)

site/networking.xml
```diff
@@ -0,0 +1,698 @@
...
+        may result in false positive and are not recommended.
+      </p>
+    </doc:section>
+
+    <doc:section name="nettick">
+      <doc:heading>Connection Handshake Timeout</doc:heading>
+
+      <p>
+        RabbitMQ has a timeout for connection handshake, 10 seconds by
+        default. When clients run in heavy constrainted environments,
```
dumbbell: "constrained"

site/production-checklist.xml
```diff
@@ -0,0 +1,291 @@
...
+        </p>
+      </doc:subsection>
+    </doc:section>
+
+
+    <doc:section name="resource-limits">
+      <doc:heading>Resource Limits</doc:heading>
+
+      <p>
+        RabbitMQ uses <a href="/alarms.html">Resource-drive alarms</a>
```
dumbbell: "Resource-driven"?

site/production-checklist.xml
```diff
@@ -0,0 +1,291 @@
...
+      </p>
+
+      <doc:subsection name="resource-limits-ram">
+        <doc:heading>Memory</doc:heading>
+        <p>
+          By default, RabbitMQ will use up to 40% of available RAM. For nodes
+          that are dedicated to run RabbitMQ, it is often reasonable to raise
+          the limit. However, care should be taken as the OS and file system caches
+          need RAM to operate as well. Failing to do so will have a severe throughput drop due
+          to OS swapping, or even result in RabbitMQ process terminated by the OS.
```
dumbbell: "in **the** RabbitMQ process terminated"

site/production-checklist.xml
```diff
@@ -0,0 +1,291 @@
...
+        <p>
+          See the section on vhosts, users, and credentials above.
+        </p>
+      </doc:subsection>
+
+      <doc:subsection name="security-considerations-erlang-cookie">
+        <doc:heading>Erlang Cookie</doc:heading>
+        <p>
+          It is highly recommended to restrict <a href="/clustering.html">Erlang cookie</a>
+          access only to the users that will run RabbitMQ and tools such as <code>rabbitmqctl</code>.
```
dumbbell: It's not only recommended on Unix, it's mandatory: the Erlang VM refuses to start if the cookie is readable, writable or executable by other people than the owner, so a mode of 0400 is the minimum on Unix (it can be writable/executable by the owner).

michaelklishin:
Links added, ready for the final round.

dumbbell:
Something I didn't pay attention to before merging is the locations of the links to the new guides in the right-hand-side menu.

"Production Checklist" is listed twice:
* in "Server Documentation";
* in "Server Documentation > Alarms".

I'm going to remove the second one.

"Networking" is listed in "Server Documentation > Clustering". I think it is more general than that and is useful for single-node production too. I'm going to move it to "Server Documentation" (one level up).

dumbbell:
@michaelklishin, do you agree with the change?

michaelklishin:
@dumbbell no objections from me.

PR #29, opened 2015-06-11
On slow links the ssl_handshake_timeout may need to be increased.

gotthardp:
Note: I know you said it should put it to the TLS section, but there is only a general introduction, which is *before* to the timeout section. I thought it may be better to talk about TLS handshake afterwards.


michaelklishin:
Thank you.

PR #30, opened 2015-06-12
Add reference useful environment variable USE_LONGNAME

tamsky:
`RABBITMQ_USE_LONGNAME` documentation
https://github.com/rabbitmq/rabbitmq-website/blob/master/site/configure.xml#L228
mentions EC2, but there's no mention of it in the ec2.xml.   This fixes that.


michaelklishin:
Please submit the same change against the `live` branch, this way we can deploy it before 3.6.0 ships ;)

PR #32, opened 2015-06-22
Preliminary docs for rabbitmq/rabbitmq-server#121

michaelklishin:
DO NOT MERGE. Pushed for a review. Subject to change.


videlalvaro:
Seems fine to me, except the <ul /> section which needs a bit of improvement, mostly around wording and cohesion

PR #38, opened 2015-07-22
fix Java sample - access to properties

arnostv:
Use properties.getContentType() instead of properties.contentType

https://www.rabbitmq.com/releases/rabbitmq-java-client/current-javadoc/com/rabbitmq/client/AMQP.BasicProperties.html#getContentType%28%29

properties.contentType is private field and is not accessible this
way in Java.

Note:
I did not submit Contributor Agreement but I assume that this change is trivial so Contributor Agreement is not required.


michaelklishin:
Thank you. Indeed, this change is too small to warrant a signed CA.

PR #44, opened 2015-08-03
fixed typo

priviterag:
fixed typo reported in https://groups.google.com/forum/#!topic/rabbitmq-users/7N4KgPVerhc


michaelklishin:
Note that this probably was intended for `live`. I will cherry-pick.

PR #60, opened 2015-08-12
Fix wording

priviterag:


michaelklishin:
Submitted against master, not live.

PR #62, opened 2015-08-14
Added more 'Related Links' for Erlang client

gotthardp:
Sometime it's not easy (for me) to find a link to the right
Erlang client documentation. I'd add more links to the website
so that I don't search so long next time.


michaelklishin:
Thank you.

PR #63, opened 2015-08-17
Use named arguments

priviterag:


michaelklishin:
This uses the code style that came with an external contribution we've gotten rid of recently, e.g. in https://github.com/rabbitmq/rabbitmq-tutorials/commit/6704d454e400caf7194441a108e7832257e6815d.

Please remove unnecessary whitespace in argument lists, e.g.

``` csharp
( string[] args )
```

should be

``` csharp
(string[] args)
```

Issue #75, opened 2015-09-13
Link to rabbitmq-codegen repository are broken


lvjp:
Because link written into [/site/java-client.xml line 168](https://github.com/rabbitmq/rabbitmq-website/blob/master/site/java-client.xml#L168) end with an unwanted /, created links are broken.

* Wrong web page `https://www.rabbitmq.com/java-client.html`
    * Actually: `git clone https://github.com/rabbitmq/rabbitmq-codegen/.git`
    * Good one: `git clone https://github.com/rabbitmq/rabbitmq-codegen.git`

PR #76, opened 2015-09-13
Repair link to Java client GitHub repository

lvjp:
Resolve issue #75.

Remove last slash in url which break link generation.


michaelklishin:
Thank you!

PR #82, opened 2015-09-28
Add missing prerequisites

### LINKED ISSUE #81: Missing requirements in Server Build Instructions
http://www.rabbitmq.com/build-server.html is missing zip and unzip as a requirement.

Also the requirement for Python 2 is not stated while Python 3 is unsupported.

Lartza:
Add missing zip and unzip requirements and specify Python version. Fixes #81


michaelklishin:
Thank you.

PR #83, opened 2015-09-29
AMQPStreamConnection class is deprecated.

jeffreydwalter:
Updated php tutorials to use AMQPConnection class instead of the deprecated AMQPStreamConnection class. It has been officially deprecated since Nov. 2014, see: https://github.com/videlalvaro/php-amqplib/commit/64eb289c9f5721daa2bd3ad0f333c5b738580843#diff-d22541aa1646efa4a61d570b8a0a0ff0

### DIFF
--- a/site/tutorials/tutorial-five-php.md
+++ b/site/tutorials/tutorial-five-php.md
@@ -159,10 +159,10 @@ The code for `emit_log_topic.php`:
     <?php
 
     require_once __DIR__ . '/vendor/autoload.php';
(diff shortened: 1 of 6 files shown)


videlalvaro:
Class deprecation happened the other way around.

jeffreydwalter:
Oh man, duh. :) I guess that's what I get for staring at a computer screen all day. 

In that case, you guys need to update the php-amqplib demo code...

Issue #84, opened 2015-09-29
Clarify how to disable subscription TTL for the MQTT plugin


priviterag:
See https://github.com/rabbitmq/rabbitmq-mqtt/issues/26

PR #85, opened 2015-09-29
explained how to disable subscription TTL for the MQTT plugin

priviterag:
fix https://github.com/rabbitmq/rabbitmq-website/issues/84


michaelklishin:
This is not correct. If the subscription key is omitted from the config, then the plugin's own default will be used.

PR #89, opened 2015-10-06
Initial instructions for AMQP 0-9-1 dissection using Wireshark 2.0.

gotthardp:
Wireshark 2.0rc1 is scheduled for Oct 14 and the final release tentatively for Nov 18. Once rc1 is out I will announce this on the rabbitmq-user list and ask for thorough testing, so I thought it may be useful to describe its features (i.e. "what to expect"). I wrote it in a form of a website, but I leave it entirely up to you if and how it should be included.

PR #92, opened 2015-10-08
Initial instructions for AMQP 0-9-1 dissection using Wireshark 2.0.

gotthardp:
Wireshark 2.0rc1 is scheduled for Oct 14 and the final release tentatively for Nov 18. Once rc1 is out I will announce this on the rabbitmq-user list and ask for thorough testing, so I thought it may be useful to describe its features (i.e. "what to expect"). I wrote it in a form of a website, but I leave it entirely up to you if and how it should be included.

This is re-submission of #89 against the live branch.


michaelklishin:
Thank you!

PR #94, opened 2015-10-09
Fix one little mistyping on a .NET Client Example

ChristopheBoucaut:
@videlalvaro, I think it is good this time ^^' sorry for the other PR.


videlalvaro:
merged, thanks!

PR #96, opened 2015-10-22
Document absolute values in memory high watermark

dcorbacho:
Documentation for https://github.com/rabbitmq/rabbitmq-server/issues/207

### DIFF
--- a/site/memory.xml
+++ b/site/memory.xml
@@ -80,6 +80,18 @@ limitations under the License.
       </p>
 
       <p>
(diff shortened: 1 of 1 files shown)


michaelklishin:
@dcorbacho FYI, the `stable` branch in this repo is rarely used: `live` is for things that can be deployed immediately, and `master` is for the next major release. We should use it more, though, and minor new features like `server#207` belong to it.

PR #98, opened 2015-10-22
Correct docs per rabbitmq/rabbitmq-server#369

michaelklishin:
Part of https://github.com/rabbitmq/rabbitmq-server/issues/369.


videlalvaro:
The docs should mention that we can configure it using the `x-queue-master-locator` argument, or the `queue-master-locator` policy.

videlalvaro:
@michaelklishin have you seen my comment here?

PR #100, opened 2015-10-31
Fix client name

jlesech:
This is the .NET/C# client, and not Java client.


michaelklishin:
@jlesech thank you. Can you please submit this against the `live` branch so that we can deploy it immediately?

PR #101, opened 2015-10-31
Fix client name

jlesech:
This is the .NET/C# client, and not Java client.


michaelklishin:
@jlesech this would merge master (which has docs for unreleased features) into `live`. I will take care of it, thank you for your time!

jlesech:
Thank you!

PR #103, opened 2015-11-02
Update java client build instructions

### LINKED ISSUE #102: Java Client build instruction are broken
After 0d1b2bd0e84b8f77b2f62d4ee90983458baef02d, the expected path to codegen directory changed from `rabbitmq-codegen` to `rabbitmq_codegen`. As commit message mentions erlang.mk layout, I assume that it is the docs that need updating.

binarin:
They've become invalid after 0d1b2bd0e84b8f77b2f62d4ee90983458baef02d to
rabbitmq-java-client, which mentions that erlang.mk-layout is now used
by default.

Fixes #102

### DIFF
--- a/site/build-java-client.xml
+++ b/site/build-java-client.xml
@@ -37,7 +37,7 @@ limitations under the License.
 	  </li>
 
 	  <li>Check the code out directly from <a href="github.html">our GitHub repositories</a>:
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thanks.

Note: this is **only** relevant for master because `live` covers released versions with the legacy build system.

PR #105, opened 2015-11-07
fix type in basic.cancel docstring

deadtrickster:


michaelklishin:
Thanks. I will cherry-pick to the 'live' branch in a bit.

PR #106, opened 2015-11-09
Fix error returned by second channel.open

deadtrickster:
It is actually `command-invalid` https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_channel.erl#L858
XML from amqp.org also states `channel-error`, so I'm not sure whether this intended and documentation should be fixed (and probably noted [here](https://github.com/rabbitmq/rabbitmq-website/blob/master/site/specification.xml)) or this is actually implementation bug and fix should be made in https://github.com/rabbitmq/rabbitmq-common

### DIFF
--- a/site/resources/specs/amqp0-9-1.extended.xml
+++ b/site/resources/specs/amqp0-9-1.extended.xml
@@ -943,7 +943,7 @@
       <doc>
         This method opens a channel to the server.
       </doc>
-      <rule name = "state" on-failure = "channel-error">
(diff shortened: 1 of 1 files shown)


michaelklishin:
If the document says it should be `CHANNEL_ERROR`, I see now reason to not switch to it: the operation isn't recoverable from and most clients would throw an exception that apps see as entirely generic and fatal. So please submit a rabbit_common change.

It would take some digging to see what tests may be asserting on this.

PR #107, opened 2015-11-14
Fix typos in uri-spec.xml

deadtrickster:


michaelklishin:
@deadtrickster that's a bit too concise. I'm not sure what the typos are exactly. You seem to be re-arranging table cells: can you please elaborate?

deadtrickster:
@michaelklishin sure, I'm implementing connection string support now and therefor writing parser tests (https://github.com/cl-rabbit/cl-bunny/blob/master/t/unit/connection-spec.lisp). I'm using uri-spec as a reference. And Appendix A tables says `""amqp://host/"` should be parsed with port="host" as well as `"amqp://host/%2f"`. Maybe I'm mistaking here but looks like a typo to me

michaelklishin:
Thank you.

PR #108, opened 2015-11-24
Changed documented default heartbeat timeout from 580 to 60 (new default as of release 3.5.5)

vitaly-krugl:


michaelklishin:
[Live](http://rabbitmq.com/heartbeats.html).

PR #109, opened 2015-11-25
Clarify what happens after setting TTL policy

binarin:
To make it clear that server resources will be freed, and not to require
anybody to dive into the source to confirm this )

### DIFF
--- a/site/ttl.xml
+++ b/site/ttl.xml
@@ -93,6 +93,13 @@ channel.queueDeclare("myqueue", false, false, false, args);</pre>
       </p>
 
       <p>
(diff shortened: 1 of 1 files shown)


michaelklishin:
@binarin please submit the same against the `live` branch, too. Thank you!

binarin:
Should it be cherry-picked to `live` or it should be done some other way?

michaelklishin:
@binarin doesn't matter, whichever you prefer.

michaelklishin:
@binarin in fact, nevermind, I'll do the cherry-picking.

PR #114, opened 2015-12-04
HIPE recomendation updated.

hairyhum:
Consider HiPE on erlang 17.x to be stable. https://github.com/rabbitmq/rabbitmq-server/issues/58


michaelklishin:
@hairyhum doc changes that can go live immediately (e.g. they don't cover an unreleased feature) should be merged into the `live` branch. `stable` is for pending patch releases, `master` is for pending feature release.

I will cherry-pick to live this time.

PR #119, opened 2015-12-10
Advanced mirrored queues configuration options

bogdando:
* Specify that the page consist of two parts:
 - A/P configuration and A/A with mirrored queues using Pacemaker OCF RA
* Describe the /usr/lib/ocf/resource.d/rabbitmq/set_rabbitmq_policy.sh
  file for advanced configuration of mirrored queues
* Do not use interleaved clones in examples as there is no need for that

Signed-off-by: Bogdan Dobrelya <bdobrelia@mirantis.com>

### DIFF
--- a/site/pacemaker.xml
+++ b/site/pacemaker.xml
@@ -24,7 +24,7 @@ limitations under the License.
   </head>
   <body>
     <p class="warning">
-      This page documents a technique for achieving
+      This part documents a technique for achieving
       active-passive high availability with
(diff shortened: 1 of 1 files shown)


bogdando:
WIP -1

bogdando:
works for me now :)

michaelklishin:
I assume this is for changes that will be in `3.5.7`? Then it needs to go into `stable` as we can't deploy it immediately.

bogdando:
yes, ok

PR #134, opened 2015-12-22
Fix links to plugin README files.

gotthardp:
There is a bad link for example at the end of http://www.rabbitmq.com/plugins.html


michaelklishin:
:+1:

PR #148, opened 2016-01-27
Fixed web-stomp documentation

matteprl:
Parts related to WebSockets support were misleading.
Special thanks to Narūnas K from StackOverflow:
http://stackoverflow.com/questions/34968313/rabbitmq-websocket-404-not-found

### DIFF
--- a/site/web-stomp.md
+++ b/site/web-stomp.md
@@ -55,9 +55,9 @@ is included as part of RabbitMQ Web STOMP examples.
 
 By default the Web STOMP plugin exposes both a WebSocket and a
 SockJS endpoint on port 15674. The WebSocket endpoint is available
-on the `/ws` path:
+on the `/stomp/websocket` path:
 
-    http://127.0.0.1:15674/ws
+    ws://127.0.0.1:15674/stomp/websocket
 
 The SockJS endpoint on the `/stomp` prefix:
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
The link above covers RabbitMQ 3.3.5 and uses Nginx. Sorry but the correct path [seems to be /ws](https://github.com/rabbitmq/rabbitmq-web-stomp-examples/blob/master/priv/bunny.html#L112) in `3.6.0`.

michaelklishin:
@essen am I missing something?

matteprl:
@michaelklishin Stackoverflow indeed covers the nginx in question. But the problem is that nginx has nothing to do with it. It couldn't work this way - as rabbitmq doesn't have anything at `/ws` path. See very bottom of question:

> EDIT
>
>RabbitMQ docs are misleading. Correct WebSocket URI:

michaelklishin:
@matteprl [yes it does](https://github.com/rabbitmq/rabbitmq-web-stomp-examples/blob/master/priv/bunny.html#L112), as of 3.6.0 there is a "raw" WebSocket endpoint at `/ws` (and SockJS emulation is at `/stomp`).

michaelklishin:
Hm, so WebSocket endpoint tests use [/stomp/websocket](https://github.com/rabbitmq/rabbitmq-web-stomp/blob/master/test/src/rabbit_ws_test_raw_websocket.erl#L40) and rabbitmq-web-stomp-examples uses `/ws`.

@essen is this intentional?

essen:
There's two "raw" Websocket: the one from /stomp, and the one at /ws (after 3.6.0). The one going through stomp doesn't support binary data (because of sockjs), the one at /ws does.

essen:
The only one documented for 3.6 should be /ws.

matteprl:
@essen That clarifies a lot. Sorry for causing trouble.

michaelklishin:
@matteprl thank you for taking the time!

PR #151, opened 2016-02-10
Fix code typo in tutorial-one-elixir

larrylv:


larrylv:
This is the correct code link https://github.com/rabbitmq/rabbitmq-tutorials/blob/5f4af537746241af6813854f1f88fde8a3cff470/elixir/receive.exs#L4

PR #161, opened 2016-02-24
Update documentation to cover both new and classic config formats

michaelklishin:
Updates the docs for http://github.com/rabbitmq/rabbitmq-server/issues/550.

### DIFF
--- a/site/access-control.xml
+++ b/site/access-control.xml
@@ -100,7 +100,7 @@ limitations under the License.
 loopback_users = none
 </pre>
 
-        Or, in the <a href="/configure.html">classic config file format</a> (<code>rabbitmq.config</code>):
+        Or, in the <a href="/configure.html#erlang-term-config-file">classic config file format</a> (<code>rabbitmq.config</code>):
 
         <pre class="sourcecode">
 [{rabbit, [{loopback_users, []}]}].</pre>
@@ -291,7 +291,7 @@ loopback_users = none
 auth_backends.1 = internal
 </pre>
 
-          Or, in the <a href="/configure.html">classic config format</a>:
+          Or, in the <a href="/configure.html#erlang-term-config-file">classic config format</a>:
 
 <pre class="example">
 [{rabbit, [
@@ -324,7 +324,7 @@ auth_backends.1 = internal
 auth_backends.1 = ldap
 </pre>
 
-          Or, in the <a href="/configure.html">classic config format</a>:
+          Or, in the <a href="/configure.html#erlang-term-config-file">classic config format</a>:
 
 <pre class="example">
 [{rabbit, [
@@ -343,7 +343,7 @@ auth_backends.1 = ldap
 auth_backends.2 = internal
 </pre>
 
-          Or, in the <a href="/configure.html">classic config format</a>:
+          Or, in the <a href="/configure.html#erlang-term-config-file">classic config format</a>:
 
       <pre class="example">
 [{rabbit, [
@@ -371,7 +371,7 @@ http.vhost_path = http://my-authenticator-app/auth/vhost
 http.resource_path = http://my-authenticator-app/auth/resource
 </pre>
 
-          Or, in the <a href="/configure.html">classic config format</a>:
+          Or, in the <a href="/configure.html#erlang-term-config-file">classic config format</a>:
 
       <pre class="example">
 [{rabbit, [
@@ -400,7 +400,7 @@ auth_backends.1.authn = internal
 auth_backends.1.authz = rabbit_auth_backend_ip_range
 </pre>
 
-          Or, in the <a href="/configure.html">classic config format</a>:
+          Or, in the <a href="/configure.html#erlang-term-config-file">classic config format</a>:
 
 <pre class="example">
 [{rabbit, [
@@ -422,7 +422,7 @@ auth_backends.1.authn = ldap
 auth_backends.1.authz = internal
 </pre>
 
-          Or, in the <a href="/configure.html">classic config format</a>:
+          Or, in the <a href="/configure.html#erlang-term-config-file">classic config format</a>:
 
 <pre class="example">
 [{rabbit, [
@@ -448,7 +448,7 @@ auth_backends.1.authz = internal
 auth_backends.2       = internal
 </pre>
 
-          Or, in the <a href="/configure.html">classic config format</a>:
+          Or, in the <a href="/configure.html#erlang-term-config-file">classic config format</a>:
 
 <pre class="example">
 [{rabbit, [
--- a/site/access-control.xml
+++ b/site/access-control.xml
@@ -453,7 +453,7 @@ auth_backends.2       = internal
 <pre class="example">
 [{rabbit, [
             {auth_backends, [{rabbit_auth_backend_ldap, rabbit_auth_backend_internal},
-                             rabbit_auth_backend_internal]
+                             rabbit_auth_backend_internal]}
           ]
  }].
 </pre>
--- a/site/access-control.xml
+++ b/site/access-control.xml
@@ -305,11 +305,11 @@ auth_backends.1 = internal
             The following aliases are available:
 
             <ul>
-              <li><code>internal</code>for <code>rabbit_auth_backend_internal</code></li>
-              <li><code>ldap</code>for <code>rabbit_auth_backend_ldap</code> (from the <a href="/ldap.html">LDAP plugin</a>)</li>
-              <li><code>http</code>for <code>rabbit_auth_backend_http</code> (from the <a href="https://github.com/rabbitmq/rabbitmq-auth-backend-http">HTTP auth backend plugin</a>)</li>
-              <li><code>amqp</code>for <code>rabbit_auth_backend_amqp</code> (from the <a href="https://github.com/rabbitmq/rabbitmq-auth-backend-amqp">AMQP 0-9-1 auth backend plugin</a>)</li>
-              <li><code>dummy</code>for <code>rabbit_auth_backend_dummy</code></li>
+              <li><code>internal</code> for <code>rabbit_auth_backend_internal</code></li>
+              <li><code>ldap</code> for <code>rabbit_auth_backend_ldap</code> (from the <a href="/ldap.html">LDAP plugin</a>)</li>
+              <li><code>http</code> for <code>rabbit_auth_backend_http</code> (from the <a href="https://github.com/rabbitmq/rabbitmq-auth-backend-http">HTTP auth backend plugin</a>)</li>
+              <li><code>amqp</code> for <code>rabbit_auth_backend_amqp</code> (from the <a href="https://github.com/rabbitmq/rabbitmq-auth-backend-amqp">AMQP 0-9-1 auth backend plugin</a>)</li>
+              <li><code>dummy</code> for <code>rabbit_auth_backend_dummy</code></li>
             </ul>
 
             When using third party plugins, providing a full module name is necessary.
--- a/site/configure.xml
+++ b/site/configure.xml
@@ -462,10 +462,10 @@ CONFIG_FILE=/etc/rabbitmq/testdir/bunnies</pre>
         </div>
       </doc:section>
 
-      <doc:section name="configuration-file">
+      <doc:section name="configuration-files">
         <doc:heading>Configuration File(s)</doc:heading>
 
-        <doc:subsection name="config-file">
+        <doc:subsection name="config-file-intro">
           <doc:heading>Introduction</doc:heading>
           <p>
             While some settings in RabbitMQ can be tuned using environment variables,
@@ -476,7 +476,7 @@ CONFIG_FILE=/etc/rabbitmq/testdir/bunnies</pre>
           </p>
         </doc:subsection>
 
-        <doc:subsection name="config-file">
+        <doc:subsection name="config-file-formats">
           <doc:heading>The New and Old Config File Formats</doc:heading>
           <p>
             Prior to RabbitMQ 3.7.0, RabbitMQ config file was named
--- a/site/configure.xml
+++ b/site/configure.xml
@@ -1067,7 +1067,7 @@ cluster_nodes.disc.2 = rabbit@hostname2
                 <li><code>pause_minority</code></li>
                 <li><code>pause_if_all_down</code></li>
               </ul>
-              <code>pause_if_all_down</code> mode require additional parameters:
+              <code>pause_if_all_down</code> mode requires additional parameters:
               <ul>
                 <li><code>nodes</code></li>
                 <li><code>recover</code></li>
--- a/site/configure.xml
+++ b/site/configure.xml
@@ -528,7 +528,11 @@ listeners.tcp.default = 5673
           AMQP 0-9-1 client connections from 5672 to 5673.
         </p>
         <p>
-           This configuration file is not the same as the environment
+          To override main RabbitMQ config file location, use the <span class="envvar">RABBITMQ_CONFIG_FILE</span>
+          environment variable.
+        </p>        
+        <p>
+           Note that this configuration file is not the same as the environment
            configuration file, <code>rabbitmq-env.conf</code>, which
            can be used to set environment variables on non-Windows
            systems.
@@ -559,6 +563,10 @@ listeners.tcp.default = 5673
             example advanced.config file</a> named
             <code>advanced.config.example</code>. It focuses on the
             options that are typically set using the advanced config.
+          </p>
+          <p>
+            To override advanced config file location, use the <span class="envvar">RABBITMQ_ADVANCED_CONFIG_FILE</span>
+            environment variable.
           </p>
         </doc:subsection>
 
--- a/site/configure.xml
+++ b/site/configure.xml
@@ -44,7 +44,7 @@ limitations under the License.
               set in the environment configuration file, <code>rabbitmq-env.conf</code>)
           </dd>
           <dt>
-              <a href="#configuration-file">One of more configuration file(s)</a>
+              <a href="#configuration-file">One or more configuration file(s)</a>
           </dt>
           <dd>
               defines server component settings for permissions, limits and
--- a/site/configure.xml
+++ b/site/configure.xml
@@ -1126,8 +1126,9 @@ cluster_nodes.disc.2 = rabbit@hostname2
         </table>
 
         <p>
-          Following configuration values can be set in 
-          <a href="#advanced-config-file">advanced config file</a> only.
+          The following configuration settings can be set in 
+          the <a href="#advanced-config-file">advanced config file</a> only,
+          under the <code>rabbit</code> section.
         </p>
 
         <table>
@@ -1171,11 +1172,11 @@ cluster_nodes.disc.2 = rabbit@hostname2
             </td>
           </tr>
           <tr>
-            <td><code>queue_index_max_journal_entries </code></td>
+            <td><code>queue_index_max_journal_entries</code></td>
             <td>
-              Tunable value for the persister. You almost certainly
-              should not change this.
-              <p>Default: <code>65536</code></p>
+              After how many queue index journal entries it will be
+              flushed to disk.
+              <p>Default: <code>32768</code></p>
             </td>
           </tr>
         </table>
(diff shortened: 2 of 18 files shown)


site/access-control.xml
```diff
@@ -297,18 +337,45 @@ limitations under the License.
...
+# rabbitmq.conf
+#
+auth_backends.1 = ldap
+# uses module name instead of a short alias, "http"
+auth_backends.2 = rabbit_auth_backend_http
+</pre>
+<pre class="example">
+%% advanced.config
+%% See HTTP backend docs for details
+{rabbitmq_auth_backend_http,
```
hairyhum: `http.user_path`
`http.vhost_path`
`http.resource_path`
are defined in http backend plugin schema https://github.com/rabbitmq/rabbitmq-auth-backend-http/commit/571937f034fd4829ca2dfdb732de87df387c43f1

site/configure.xml
```diff
@@ -421,10 +495,56 @@ CONFIG_FILE=/etc/rabbitmq/testdir/bunnies</pre>
            can be used to set environment variables on non-Windows
            systems.
         </p>
+        <p>
+            RabbitMQ server source repository contains
+            <a href="https://github.com/rabbitmq/rabbitmq-server/blob/stable/docs/rabbitmq.conf.example">an example rabbitmq.conf file</a>
+            named <code>rabbitmq.config.example</code>. This
```
hairyhum: Should it be `rabbitmq.conf.example`?

site/configure.xml
```diff
@@ -421,10 +495,56 @@ CONFIG_FILE=/etc/rabbitmq/testdir/bunnies</pre>
...
+            Some configuration settings are not possible or difficult to configure
+            using the sysctl format. As such, it is possible to use an additional
+            config file in the Erlang term format (same as <code>rabbitmq.config</code>).
+            That file is commonly named <code>advanced.config</code>. It will be merged
+            with the configuration provided in <code>rabbitmq.conf</code>.
+          </p>
+          <p>
+            RabbitMQ server source repository contains
+            <a href="https://github.com/rabbitmq/rabbitmq-server/blob/stable/docs/advanced.config.example">an example advanced.config file</a>
+            named <code>rabbitmq.config.example</code>. This
```
hairyhum: Same as above comment. Should be `advanced.config.example`

site/configure.xml
```diff
@@ -497,60 +621,62 @@ CONFIG_FILE=/etc/rabbitmq/testdir/bunnies</pre>
...
           </tr>
+          
           <tr>
             <td><code>ssl_options</code></td>
             <td>
-              SSL configuration. See the <a href="ssl.html#enabling-ssl">SSL
+              TLS configuration. See the <a href="ssl.html#enabling-ssl">TLS support
               documentation</a>.
-              <p>Default: <code>[]</code></p>
+              <p>Default: <code>none</code></p>
```
hairyhum: If `none` is defined, none of other options will work.
For example:
```
ssl_options = none
ssl_options.verify = verify_peer
```
Equivalent to 
```
ssl_options = none
```

michaelklishin: The only idea I have is to set "undefined", which can lead to exactly the same problem. Yeah, I don't think many will fall into this trap.

hairyhum:
@michaelklishin: do we want `auth_backends`, `auth_mechanisms` and `cluster_nodes` to have `none` option?

michaelklishin:
@hairyhum I think it would be helpful in case someone uncomments/uses the default value unintentionally.

michaelklishin:
for `auth_backends` and `auth_mechanisms`, having no value will reject all client connections, so that probably should fail validation.

hairyhum:
So it's better to leave it without `none`

site/configure.xml
```diff
@@ -44,7 +44,7 @@ limitations under the License.
               set in the environment configuration file, <code>rabbitmq-env.conf</code>)
           </dd>
           <dt>
-              <a href="#configuration-file">a configuration file</a>
+              <a href="#configuration-file">One of more configuration file(s)</a>
```
camelpunch: Should this be "One or more", or maybe "One of many"?

site/configure.xml
```diff
@@ -391,20 +393,92 @@ CONFIG_FILE=/etc/rabbitmq/testdir/bunnies</pre>
...
+          <p>
+            While the new config format is more convenient for humans to edit
+            and machines to generate, it is also relatively limited compared
+            to the classic config format used prior to RabbitMQ 3.7.0. For
+            example, when configuring <a href="/ldap.html">LDAP support</a>,
+            it may be necessary to use deeply nested data structures to
+            express desired configuration. To accommodate this need,
+            RabbitMQ still supports the classic <code>rabbitmq.config</code>
+            config files as well as ability to use both formats at the same time
+            (<code>advanced.config</code>). This is covered in more details in the following sections.
```
camelpunch: "more detail"

site/access-control.xml
```diff
@@ -92,11 +92,17 @@ limitations under the License.
...
-          <code>[]</code>. A complete <code>rabbitmq.config</code>
+          <code>loopback_users</code> configuration to
+          <code>none</code>. A complete <a href="/configure.html">RabbitMQ config file</a>
           which does this would look like:
         </p>
         <pre class="sourcecode">
+loopback_users = none
+</pre>
+
+        Or, in the <a href="/configure.html#erlang-term-config-file">classic config file format</a> (<code>rabbitmq.config</code>):
```
dumbbell: I would use "legacy" instead of "classic" to designate the old file formar.

site/access-control.xml
```diff
@@ -297,18 +337,42 @@ limitations under the License.
...
+# rabbitmq.conf
+#
+auth_backends.1 = ldap
+# uses module name instead of a short alias, "http"
+auth_backends.2 = rabbit_auth_backend_http
+
+# See HTTP backend docs for details
+http.user_path = http://my-authenticator-app/auth/user
+http.vhost_path = http://my-authenticator-app/auth/vhost
+http.resource_path = http://my-authenticator-app/auth/resource
```
dumbbell: I find the `http` prefix to be too generic in the context of the authentication backend. Maybe it should be the full module name? Or an additional prefix before the authentication mechanism name?

site/access-control.xml
```diff
@@ -328,7 +391,17 @@ limitations under the License.
...
             IP range backend</a> for authorisation:
           </p>
-            
+
+<pre class="example">
+# rabbitmq.conf
+#
+auth_backends.1.authn = internal
+# uses module name because this backend is from a 3rd party
+auth_backends.1.authz = rabbit_auth_backend_ip_range
```
dumbbell: I don't know enough how authentication and authorization are handled by RabbitMQ but this configuration parameters look reversed.

I would expect something like:
```
auth_backends.authn.1 = internal
auth_backends.authn.2 = http
auth_backends.authz.1 = rabbit_auth_backend_ip_range
```
In other words: RabbitMQ checks authentication first using the listed modules, then it checks authorization using another set of modules. Of course, correct me if my understanding of RabbitMQ internals is wrong.

site/access-control.xml
```diff
@@ -358,16 +441,22 @@ limitations under the License.
       </p>
 
 <pre class="example">
+# rabbitmq.conf
+#
+auth_backends.1.authn = ldap
+auth_backends.1.authz = internal
+auth_backends.2       = internal
```
dumbbell: Ok, so disregard my previous comment, because authentication and authorization backends are organized as pairs. Sorry for the noise!

site/access-control.xml
michaelklishin: We still support this format, and it is still used in the advanced config. It would be odd to call something legacy and then use it for both backwards compatibility and advanced cases.

site/access-control.xml
michaelklishin: I agree. Perhaps we should use module names instead (and not only in the docs).

hairyhum: Use `auth_http` and `auth_ldap` instead `http` and `ldap` in plugin settings?

michaelklishin: `auth_http` and `auth_ldap` sound good to me.

dumbbell: `auth_http` is still too confusing IMHO. It could be considered a parameter of the management plugin. Using the module name makes it clear we are talking about the authentication backend.

About aliases in eg. `auth_backends.1 = http`, it's ok because it's in the context of `auth_backends`.

site/access-control.xml
```diff
@@ -282,10 +301,31 @@ limitations under the License.
...
           <p>
+            The example above uses an alias, <code>internal</code> for <code>rabbit_auth_backend_internal</code>.
+            The following aliases are available:
+
+            <ul>
+              <li><code>internal</code> for <code>rabbit_auth_backend_internal</code></li>
+              <li><code>ldap</code> for <code>rabbit_auth_backend_ldap</code> (from the <a href="/ldap.html">LDAP plugin</a>)</li>
+              <li><code>http</code> for <code>rabbit_auth_backend_http</code> (from the <a href="https://github.com/rabbitmq/rabbitmq-auth-backend-http">HTTP auth backend plugin</a>)</li>
+              <li><code>amqp</code> for <code>rabbit_auth_backend_amqp</code> (from the <a href="https://github.com/rabbitmq/rabbitmq-auth-backend-amqp">AMQP 0-9-1 auth backend plugin</a>)</li>
+              <li><code>dummy</code> for <code>rabbit_auth_backend_dummy</code></li>
```
dumbbell: How are aliases managed? Is it a hard-coded list? Or does RabbitMQ try to use `rabbit_auth_backend_${alias}`?

michaelklishin: A hard-coded list.

site/access-control.xml
michaelklishin: [See this internals guide](https://github.com/rabbitmq/internals/blob/master/authorization_and_authentication.md).

site/access-control.xml
michaelklishin: Unfortunately if we go with module names in some places but not others, it will be more confusing than using just module names. Which is what we will do now.

site/configure.xml
```diff
@@ -254,6 +257,75 @@ CONFIG_FILE=/etc/rabbitmq/testdir/bunnies</pre>
...
+                    <b>Mac OS X (Homebrew)</b> - <span class="path"><span
+                    class="envvar">${install_prefix}</span>/etc/rabbitmq/rabbitmq</span>,
+                    the Homebrew prefix is usually <code>/usr/local</code>
+                  </li>
+                  <li><b>Windows</b> - <span class="path"><span class="envvar">%APPDATA%</span>\RabbitMQ\rabbitmq</span></li>
+                </ul>
+              </td>
+              <td>
+                Main RabbitMQ config file path without the <code>.conf</code>
+                (or <code>.config</code>, for the classic format) extension.
```
dumbbell: This "without the extension" thing is so confusing. I know RabbitMQ does this for a long time, but can't we document that we require the full filename but still support the missing suffix internally?

The fact that Erlang takes a filename without its `.config` extension is a technical detail we don't need to expose to end users.

michaelklishin: @dumbbell see https://github.com/rabbitmq/rabbitmq-server/issues/691.

site/configure.xml
```diff
@@ -484,11 +690,10 @@ CONFIG_FILE=/etc/rabbitmq/testdir/bunnies</pre>
         </doc:subsection>
 
         <doc:subsection name="config-items">
-        <doc:heading>Variables Configurable in rabbitmq.config</doc:heading>
+        <doc:heading>Commonly Used Variables Configurable in rabbitmq.conf</doc:heading>
         <p>
-          Many users of RabbitMQ never need to change any of these
-          values, and some are fairly obscure. However, for
-          completeness they are all listed here.
+          These values are configured most often. The list is not complete, as
```
dumbbell: This sentence is unclear. What do you think of:
> These variables are the most common.

site/configure.xml
dumbbell: To be consistent with other examples, you should recall the parameter:
```html
<p>Default: <code>ssl_options = none</code></p>
```

site/configure.xml
```diff
@@ -497,71 +702,87 @@ CONFIG_FILE=/etc/rabbitmq/testdir/bunnies</pre>
...
+              details and examples.
+              <p>Default: <code>listeners.ssl = none</code></p>
             </td>
           </tr>
           <tr>
-            <td><code>num_ssl_acceptors</code></td>
+            <td><code>num_acceptors.ssl</code></td>
             <td>
-              Number of Erlang processes that will accept connections for the SSL
+              Number of Erlang processes that will accept connections for the TLS
```
dumbbell: By the way, that's a good idea to replace SSL with TLS everywhere!

site/configure.xml
```diff
@@ -883,17 +1061,19 @@ CONFIG_FILE=/etc/rabbitmq/testdir/bunnies</pre>
...
-                <li><code>pause_minority</code></li>
-                <li><code>{pause_if_all_down, [nodes], ignore | autoheal}</code>
-                  where <code>[nodes]</code> is a list of node names<br/>
-                  (ex: <code>['rabbit@node1', 'rabbit@node2']</code>)
-                </li>
                 <li><code>autoheal</code></li>
+                <li><code>pause_minority</code></li>
+                <li><code>pause_if_all_down</code></li>
+              </ul>
+              <code>pause_if_all_down</code> mode require additional parameters:
```
dumbbell: s/require/requires/

site/disk-alarms.xml
```diff
@@ -83,12 +83,18 @@ Disabling disk free space monitoring</pre>
         required to be free on the database partition. (See the description of
         <a href="/relocate.html">file locations</a> for the default location).
         This configuration file sets the disk free space limit to 1GB:
+        <pre class="sourcecode">disk_free_limit.absolute = 1000000000</pre>
+        Or you can use memory units (KB, MB GB etc.) like this:
```
dumbbell: `kB` here? (note the lowercase `k`)

hairyhum: This format uses `cuttlefish` [bytesize](https://github.com/basho/cuttlefish/wiki/Datatypes#bytesize) datatype. So it should be KB

site/management.xml
```diff
@@ -328,36 +334,30 @@ limitations under the License.
...
-            the <code>rabbitmq_management</code> application to the name
-            of a directory in which logs can be created and restart
-            RabbitMQ. Note that only requests to the API
-            at <code>/api</code> are logged, not requests to the static
-            files which make up the browser-based GUI.
+            Note that the definitions in the file will overwrite
+            anything already in the broker; using this option will not
+            delete anything that is already there. However, if you
+            start from a completely reset broker, use of this
+            option <b>will</b> prevent the usual default user /
```
dumbbell: Please use `<strong>` instead of `<b>`.

site/management.xml
```diff
@@ -328,36 +334,30 @@ limitations under the License.
...
+            virtual host / permissions from being created.
           </p>
-        </doc:subsection>
+        </doc:subsection>
```
dumbbell: There are trailing spaces remaining here.

site/configure.xml
```diff
@@ -497,71 +702,87 @@ CONFIG_FILE=/etc/rabbitmq/testdir/bunnies</pre>
...
-              List of ports on which to listen for AMQP connections
-              (without SSL). Can contain integers (meaning "listen on
-              all interfaces") or tuples such as <code>{"127.0.0.1",
-              5672}</code> to listen on one or more interfaces.
+              Ports or hostname/pair on which to listen for AMQP connections
+              (without TLS). See the <a href="/networking.html">Networking guide</a> for more
+              details and examples.
 
-              <p>Default: <code>[5672]</code></p>
+              <p>Default: <code>listeners.tcp.default = 5672</code></p>
```
dumbbell: The listeners configuration (for RabbitMQ/AMQP and the plugins too) is very confusing.

1. `tcp` or `ssl` below make no sense (I know it comes from the classic format). TLS is still over TCP and currently we support TCP or HTTP as the transport layer and this is orthognal to TLS. Can we use the protocol name or purpose instead (see example below)?
2. What is `default` in `listeners.tcp.default`?
3. If I compare the configuration for AMQP(S) and the management plugin:
 * `listeners` vs. `listener` (plural vs. singular)
 * `...tcp.default = 5672` vs. `...port = 15672`
 * `listeners.ssl = none` vs. `management.listener.ssl = true` (`none`/`true`)
 * `ssl_options` vs. `ssl_opts`

Below, I'm giving an example of a centralized configuration for listeners, with the following additional changes:
* I put the acceptors count and TCP options as leaves there as well.
* I use `tls` instead of `ssl` to reflect the naming used in the documentation.
* `listeners` may be too long; `listen` would be enough.

```
listen.amqp.port = 5672
listen.amqp.tls = true # Or "tls.enabled = true" if an intermediate node can't have a value
listen.amqp.tls.port = 5671
listen.amqp.tls.certfile = /path/to/cert.pem
listen.mqtt.port = 1883
listen.mqtt.tls = false
listen.mqtt.tls.port = 8883 # Ignored because TLS is disabled
listen.management.port = 15672
listen.management.tls = true
listen.management.tls.port = 15671

# For $protocol over HTTP:
listen.stomp.http.port = ...
listen.stomp.http.tls.port = ...
```

Default/common options could be set under a `default` node:
```
listen.default.nb_acceptors = 1
listen.default.handshake_timeout = 10000

listen.default.tls.enabled = true
listen.default.tls.certfile = /path/to/cert.pem
listen.default.tls.keyfile = /path/to/key.pem
listen.default.tls.nb_acceptors = 2 # Otherwise defaults to listen.default.nb_acceptors
```

This lacks an example of how to listen on specific interfaces.

What do you think?

michaelklishin: 1. TLS is over TCP but there is no word for "non-TLS" ("plain" is not widely used).
2. It's a name that can be anything.
3. Management UI probably nearly always has a single listener configured, so this is in part by design.
4. We could replace `none` with `false` but `none` means something else: it means "no options"
5. We discussed this before. If we change this to `ssl_options` it will confuse those who are used to the classic format, which is the vast majority of people. The best I can suggest is that we support both in the management plugin.

site/networking.xml
```diff
@@ -98,6 +104,13 @@ limitations under the License.
           to listen on localhost only for both IPv4 and IPv6:
 
 <pre class="code">
+listeners.tcp.1 = 127.0.0.1:5672
+listeners.tcp.2 = ::1:5672
```
dumbbell: Is it possible to indicate an IP address without the port (so the default port is used)?

If yes, we must enclose IPv6 addresses inside square brackets (`[::1]`) like we do in URLs.

michaelklishin: Good point.

site/disk-alarms.xml
michaelklishin: [We do support `kB`](https://github.com/rabbitmq/rabbitmq-server/blob/master/src/rabbit_resource_monitor_misc.erl#L43).

site/configure.xml
michaelklishin: Yes but not in scope for this issue. I do it as drive-by changes whenever I work on the docs.

site/disk-alarms.xml
hairyhum: We do, but cuttlefish configuration doesn't.

site/networking.xml
```diff
@@ -105,7 +105,7 @@ Or using classic config format:
 
 <pre class="code">
 listeners.tcp.1 = 127.0.0.1:5672
-listeners.tcp.2 = ::1:5672
+listeners.tcp.2 = [::1]:5672
```
hairyhum: This should be added to example snippets in `rabbitmq_tests`

michaelklishin: I've added it and waiting for the test suite to finish.

michaelklishin: I reverted that change due to https://github.com/basho/cuttlefish/issues/216. We'll work on that later.

site/disk-alarms.xml
```diff
@@ -83,12 +83,18 @@ Disabling disk free space monitoring</pre>
         required to be free on the database partition. (See the description of
         <a href="/relocate.html">file locations</a> for the default location).
```
bdshroyer: I'm not sure from the context here what the "default location" is referring to. Is it the free disk space limit?

michaelklishin: Default database location.

PR #171, opened 2016-03-22
Use one-liners for quick copy/paste and less traces

dastergon:
### Changelog
* Moved `rabbitmq` APT repository to separate sources list
* Changed some installation commands into one-liners to decrease the number of required steps
* Minor instruction enhancements

### DIFF
--- a/site/install-debian.xml
+++ b/site/install-debian.xml
@@ -104,8 +104,10 @@ limitations under the License.
         <h3>Using rabbitmq.com APT Repository</h3>
 	<ol>
 	  <li>
-	    Add the following line to your <code>/etc/apt/sources.list</code>:
-	    <pre class="sourcecode">deb http://www.rabbitmq.com/debian/ testing main</pre>
+	    Execute the following command to add the APT repository to your <code>/etc/apt/sources.list.d</code>:
+	    <pre class="sourcecode">
+            echo 'deb http://www.rabbitmq.com/debian/ testing main' | sudo tee /etc/apt/sources.list.d/rabbitmq.list
+	    </pre>
 	    (Please note that the word <code>testing</code> in this
 	    line refers to the state of our release of RabbitMQ, not
 	    any particular Debian distribution. You can use it with
--- a/site/install-debian.xml
+++ b/site/install-debian.xml
@@ -125,7 +125,7 @@ limitations under the License.
 	  </li>
 	  <li>
 	    Run the following command to update the package list:
-	    <pre class="sourcecode">sudo apt-get update</pre>.
+	    <pre class="sourcecode">sudo apt-get update</pre>
 	  </li>
 	  <li>
 	    Install <code>rabbitmq-server</code> package:
--- a/site/install-debian.xml
+++ b/site/install-debian.xml
@@ -118,11 +118,10 @@ limitations under the License.
 	    our <a href="rabbitmq-signing-key-public.asc">public
 	      key</a> to your trusted key list using
 	    <code>apt-key(8)</code>:
-
-	    <pre class="sourcecode">wget https://www.rabbitmq.com/rabbitmq-signing-key-public.asc
-sudo apt-key add rabbitmq-signing-key-public.asc</pre>
-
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you.

@hairyhum: please QA this in master and then cherry-pick to `live`.

site/install-debian.xml
```diff
@@ -104,8 +104,10 @@ limitations under the License.
         <h3>Using rabbitmq.com APT Repository</h3>
 	<ol>
 	  <li>
-	    Add the following line to your <code>/etc/apt/sources.list</code>:
-	    <pre class="sourcecode">deb http://www.rabbitmq.com/debian/ testing main</pre>
+	    Execute the following command to add the APT repository to your <code>/etc/apt/sources.list.d</code>:
+	    <pre class="sourcecode">
+            echo 'deb http://www.rabbitmq.com/debian/ testing main' | sudo tee /etc/apt/sources.list.d/rabbitmq.list
```
hairyhum: One liners don't fit into page layout. Can be made multiline.
Also code should start right after `<pre>` tag, because leading whitespace is not being trimmed.

michaelklishin: Yes, leading whitespace in `pre` sections is something we should avoid.

dastergon: I could send an extra commit with code after `<pre>` tag but I am not really sure how the one-liners  could be handled in the page layout.

hairyhum: Inside `<pre>` tags line breaks are preserved so you should add `\` and line break in long commands.

dastergon: Is this format fine? 
```
<pre class="sourcecode"> echo 'deb http://www.rabbitmq.com/debian/ testing main' | \ 
sudo tee /etc/apt/sources.list.d/rabbitmq.list </pre>
```

hairyhum: You can even avoid using `\`, because pipeline supports multilining
```
echo 'deb http://www.rabbitmq.com/debian/ testing main' |
sudo tee /etc/apt/sources.list.d/rabbitmq.list
```

michaelklishin: @dastergon it is fine. We [do it differently](https://github.com/rabbitmq/rabbitmq-website/blob/master/site/api-guide.xml#L110) in most places but for this guide you don't have to worry about that.

dastergon: @michaelklishin @hairyhum : just sent an extra commit with the layout fixes. Let me know if there's anything else for change.

site/install-debian.xml
```diff
@@ -118,18 +119,16 @@ limitations under the License.
...
-
-            <p>Our public signing key is also <a href="https://bintray.com/user/downloadSubjectPublicKey?username=rabbitmq">available from Bintray</a>.</p>
+	    <pre class="sourcecode">wget -O- https://www.rabbitmq.com/rabbitmq-signing-key-public.asc |
+                sudo apt-key add -</pre>
+	    <p>Our public signing key is also <a href="https://bintray.com/user/downloadSubjectPublicKey?username=rabbitmq">available from Bintray</a>.</p>
 	  </li>
 	  <li>
-	    Run <code>apt-get update</code>.
+	    Run the following command to update the package list:
+	    <pre class="sourcecode">sudo apt-get update</pre>.
```
hairyhum: Full stop in the end of line should be removed.

site/install-debian.xml
```diff
@@ -104,8 +104,9 @@ limitations under the License.
         <h3>Using rabbitmq.com APT Repository</h3>
 	<ol>
 	  <li>
-	    Add the following line to your <code>/etc/apt/sources.list</code>:
-	    <pre class="sourcecode">deb http://www.rabbitmq.com/debian/ testing main</pre>
+	    Execute the following command to add the APT repository to your <code>/etc/apt/sources.list.d</code>:
+	    <pre class="sourcecode">echo 'deb http://www.rabbitmq.com/debian/ testing main' |
+                sudo tee /etc/apt/sources.list.d/rabbitmq.list</pre>
```
hairyhum: There can be less indent, since it's including all spaces from the beginning of the line.

dastergon: Fixed indents in a new commit

hairyhum:
Looks good now. Thanks, @dastergon

PR #178, opened 2016-04-06
Reuse intro text in tutorial 1 of all languages

camelpunch:
Use an XML include to reuse the introductory text.


camelpunch:
Closing in favour of merging into live.

PR #186, opened 2016-04-19
Difference in PerfTest topology with or without queue name

### LINKED ISSUE #185: Information about PerfTest topology
PerfTest is almost undocumented. Adding information about queues topology in different configurations can be very useful.

hairyhum:
Added some information about difference in queues and consumers with `-u queue_name` and without it.
Fixes #185
Maybe there are something more to be added?

### DIFF
--- a/site/java-tools.xml
+++ b/site/java-tools.xml
@@ -86,18 +86,18 @@ limitations under the License.
       </dl>
 
       <p>
-      If queue name is defined (<code>-u "queue-name"</code>), PerfTest will create a queue with this name
-      and all consumers will consume from this queue. The queue will be bound to the direct exchange with it's name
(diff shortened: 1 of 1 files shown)


site/java-tools.xml
```diff
@@ -84,6 +84,25 @@ limitations under the License.
           Start 10 consumers from a predeclared queue, and no producers.
         </dd>
       </dl>
+
+      <p>
+      If queue name is defined (<code>-u "queue-name"</code>), PerfTest will create a queue with this name
+      and all consumers will consume from this queue. The queue will be bound to the direct exchange with it's name
```
camelpunch: "its name" - when an it possesses something you shouldn't use an apostrophe.

PR #199, opened 2016-04-26
Fix teaser reveals in tutorial 5

camelpunch:
Use CSS targets to hide/reveal teasers.


camelpunch:
GitHub's fancy forms didn't notice I chose the live branch.

PR #209, opened 2016-05-09
Fix dead letter cycle detection note

### LINKED ISSUE #206: Incorrect documentation about cycle detection in DLX
According to docs, message is removed from cycle only **if entire cycle is due to message expiry**.

This commit https://github.com/rabbitmq/rabbitmq-server/commit/f8f665a065468201bbb42dcc5cd5143259219e34 changed it to be **if there was no `rejected` reasons in cycle**

Docs should be updated to correspond with behaviour.

hairyhum:
Fixes #206


michaelklishin:
Thank you!

PR #213, opened 2016-05-12
Replace old OpenPGP keys with a new stronger key

dumbbell:
References rabbitmq/rabbitmq-server#718.
[#118296861]


dumbbell:
The patch is ready for review. I marked the pull request as "DO NOT MERGE" because I will update the patch before the final merge with:
* the appropriate date in the news entry
* a proper (non-fake) example in `signatures.xml`, though I may do so after the release of 3.6.2 if you use the new key.

PR #226, opened 2016-06-28
Fixed tutorial-two-java

### LINKED ISSUE #204: Tutorial 2, Java: Couple of confusions in code
1) In the section "Message acknowledgment"

Its only mentioned in the "text" that autoAck flag needs to be set to false. Its not shown in the code. Its not at all clear where the flag was set in the first place. 
I suggest showing it in the code highlighting the parts that changed.
That is, from channel.basicConsume(QUEUE_NAME, true, defaultConsumer) to channel.basicConsume(QUEUE_NAME, false, defaultConsumer) or better to channel.basicConsume(QUEUE_NAME, NO_AUTO_ACK, defaultConsumer) so that its clear what the flag is for.

2) The same section in the highlighted code also asks to set, channel.basicQos(1) without introducing it. Later in the same article it seems like we use it to dispatch messages fairly amongst consumers in contrast to round robin fashion. In case my understanding is correct and channel.basicQos has nothing to do with messageAcknowledgement I'll suggest to remove that from the code snippet.

ffissore:
Tutorial two for java contains a mistake. This PR fixes it. Closes #204.

### DIFF
--- a/site/tutorials/tutorial-two-java.md
+++ b/site/tutorials/tutorial-two-java.md
@@ -464,7 +464,7 @@ And our `Worker.java`:
             }
           }
         };
-        channel.basicConsume(TASK_QUEUE_NAME, false, consumer);
+        channel.basicConsume(TASK_QUEUE_NAME, consumer);
       }
 
       private static void doWork(String task) {
--- a/site/tutorials/tutorial-two-java.md
+++ b/site/tutorials/tutorial-two-java.md
@@ -231,8 +231,8 @@ the consumer dies. It's fine even if processing a message takes a very, very
 long time.
 
 Message acknowledgments are turned on by default. In previous
-examples we explicitly turned them off via the `autoAck=true`
-flag. It's time to remove this flag and send a proper acknowledgment
+examples we explicitly turned them off passing `true` to `autoAck` flag of method
+`basicConsume`. It's time to remove this flag and send a proper acknowledgment
 from the worker, once we're done with a task.
 
     :::java
@@ -252,6 +252,7 @@ from the worker, once we're done with a task.
         }
       }
     };
+    channel.basicConsume(TASK_QUEUE_NAME, consumer);
 
 
 Using this code we can be sure that even if you kill a worker using


site/tutorials/tutorial-two-java.md
```diff
@@ -463,7 +464,7 @@ And our `Worker.java`:
             }
           }
         };
-        channel.basicConsume(TASK_QUEUE_NAME, false, consumer);
+        channel.basicConsume(TASK_QUEUE_NAME, consumer);
```
michaelklishin: Please use the method that explicitly passes in the automatic acknowledgements parameter.

ffissore: Original documentation above says "It's time to remove this flag". Should I change that as well?

michaelklishin: Ah, OK. No need to edit anything, I need to read the whole thing before making any conclusions, sorry.

michaelklishin:
@kjnilsson I keep failing to get to this, so re-assigning to you. Note that we need to keep the tutorials in sync with https://github.com/rabbitmq/rabbitmq-tutorials.

site/tutorials/tutorial-two-java.md
```diff
@@ -231,8 +231,8 @@ the consumer dies. It's fine even if processing a message takes a very, very
 long time.
 
 Message acknowledgments are turned on by default. In previous
-examples we explicitly turned them off via the `autoAck=true`
-flag. It's time to remove this flag and send a proper acknowledgment
+examples we explicitly turned them off passing `true` to `autoAck` flag of method
+`basicConsume`. It's time to remove this flag and send a proper acknowledgment
```
acogoluegnes: I tend to prefer the explicit method, so the wording would be _It's time to set this flag to `false` and ..._

site/tutorials/tutorial-two-java.md
```diff
@@ -252,6 +252,7 @@ from the worker, once we're done with a task.
         }
       }
     };
+    channel.basicConsume(TASK_QUEUE_NAME, consumer);
```
acogoluegnes: Obviously, we need to add the call in this sample. Maybe something like this:
```java
boolean autoAck = false;
channel.basicConsume(TASK_QUEUE_NAME, autoAck, consumer);
```

site/tutorials/tutorial-two-java.md
acogoluegnes: Same thing as previously:

```java
boolean autoAck = false;
channel.basicConsume(TASK_QUEUE_NAME, autoAck, consumer);
```

acogoluegnes:
I added a [few comments](https://github.com/rabbitmq/rabbitmq-website/pull/226/files)

A few more suggestions:

Change first auto-ack consumer from:
```java
channel.basicConsume(TASK_QUEUE_NAME, true, consumer);
```
to:
```java
boolean autoAck = true;
channel.basicConsume(TASK_QUEUE_NAME, autoAck, consumer);
```

As suggested by dhruvkapur91, add a note for `basicQos` on the first use:
```java
channel.basicQos(1); // accept only one unack-ed message at a time (see below)
```

michaelklishin:
We manually applied this in https://github.com/rabbitmq/rabbitmq-website/pull/248, thank you, @ffissore!

ffissore:
Glad it helped :)

PR #228, opened 2016-06-30
Removed useless 'an'

ffissore:


michaelklishin:
If you want something to be deployed soon and not when `3.7.0` ships, please submit changes against the `live` branch.

PR #231, opened 2016-07-20
use `Buffer.from` in place of `new Buffer`

Robinnnnn:
Node has [deprecated](https://nodejs.org/api/buffer.html#buffer_class_buffer) the use of `new Buffer()` as of v6.0.0 and has assigned it a stability score of 0. The Javascript tutorial should utilize [`Buffer.from`](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_str_encoding) instead, which was introduced in v5.10.0.

Specifically, the [message sending example](https://www.rabbitmq.com/tutorials/tutorial-one-javascript.html) with this line:

`ch.sendToQueue(q, new Buffer('Hello World!'));`

should be updated to:

`ch.sendToQueue(q, Buffer.from('Hello World!'));`


michaelklishin:
Please branch off of `live`. `master` will be deployed when `3.7.0` ships.

PR #232, opened 2016-07-20
Fix tutorial-one-javascript. Close #230

### LINKED ISSUE #230: Javascript Tutorial #1: new Buffer is deprecated as of Node v6.0.0
Node has [deprecated](https://nodejs.org/api/buffer.html#buffer_class_buffer) the use of `new Buffer()` as of v6.0.0 and has assigned it a stability score of 0. The Javascript tutorial should utilize [`Buffer.from`](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_str_encoding) instead, which was introduced in v5.10.0.

Specifically, the [message sending example](https://www.rabbitmq.com/tutorials/tutorial-one-javascript.html) with this line:

`ch.sendToQueue(q, new Buffer('Hello World!'));`

should be updated to:

`ch.sendToQueue(q, Buffer.from('Hello World!'));`

Robinnnnn:
Replace deprecated `new Buffer` instantiation brought up in [#230](https://github.com/rabbitmq/rabbitmq-website/issues/230).

### DIFF
--- a/site/tutorials/tutorial-one-javascript.md
+++ b/site/tutorials/tutorial-one-javascript.md
@@ -95,7 +95,7 @@ to the queue:
         var q = 'hello';
 
         ch.assertQueue(q, {durable: false});
-        ch.sendToQueue(q, new Buffer('Hello World!'));
+        ch.sendToQueue(q, Buffer.from('Hello World!'));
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you!

PR #237, opened 2016-08-09
Updating docs to reflect omitted linger statement

mixacha:
Work in progress, please take a look and I'll update additional docs (site/networking) to reflect default options as well.


mixacha:
@michaelklishin ping

michaelklishin:
@mixacha thanks, can you please also update the `live` branch? it has examples in the 3.6.x config format.

michaelklishin:
OK, so I'm not 100% sure what the value should be in the new config, so I'll undo the master change but please update `live`, which is 100% Erlang terms examples.

michaelklishin:
Yup, there is no mapping in the new config for that key. Thanks for helping us discover it :)

PR #243, opened 2016-08-17
Fix: Add missing option in example openssl.cnf

Thubo:
Add `domainComponent = optional` in the `[ testca_policy ]` section of the openssl.cnf. Otherwise ssl/tls connections will fail in some cases. 

See also:
* https://groups.google.com/forum/#!topic/rabbitmq-users/uwRLaKqpha0
* https://github.com/michaelklishin/tls-gen/

### DIFF
--- a/site/ssl.xml
+++ b/site/ssl.xml
@@ -140,6 +140,7 @@ countryName = optional
 emailAddress = optional
 organizationName = optional
 organizationalUnitName = optional
+domainComponent = optional
 
 [ certificate_extensions ]
 basicConstraints = CA:false


michaelklishin:
@Thubo I have corrected your PR's language a bit because it doesn't affect most users. Thanks. If there is a PR for tls-gen that you have in mind, I'd happily consider it!

Thubo:
@michaelklishin Thanks for merging. We need SANs for our certificates, if I find the time I will send a PR to tls-gen. It's a nice tool for quickly getting things up and running.

PR #249, opened 2016-08-24
Clarify acknowledgment in Java tutorial 2

### LINKED ISSUE #204: Tutorial 2, Java: Couple of confusions in code

acogoluegnes:
As suggested by @dhruvkapur91 and contributed by @ffissore.

Fixes #204

### DIFF
--- a/site/tutorials/tutorial-two-java.md
+++ b/site/tutorials/tutorial-two-java.md
@@ -119,7 +119,8 @@ delivered messages and perform the task, so let's call it `Worker.java`:
         }
       }
     };
-    channel.basicConsume(TASK_QUEUE_NAME, true, consumer);
+    boolean autoAck = true; // acknowledgment is covered below
+    channel.basicConsume(TASK_QUEUE_NAME, autoAck, consumer);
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
This is already live.

PR #256, opened 2016-08-30
Update handling split-brain in Pacemaker guide

dmitrymex:
Added master health check to the picture for clarity.


dmitrymex:
@bogdando : what do you think?

bogdando:
:+1:

PR #261, opened 2016-09-06
Fix little typo mistake.

jmferrer:


michaelklishin:
Thank you but it is not a typo: "eldest" seems appropriate there.

jmferrer:
Sorry.
I did not know that word but it exists: http://dictionary.cambridge.org/es/gramatica/gramatica-britanica/elder-eldest-or-older-oldest

PR #262, opened 2016-09-06
In Pacemaker guide do not suggest to run the second monitor

dmitrymex:
See the following PR for the reasoning:
https://github.com/rabbitmq/rabbitmq-server/pull/942


dmitrymex:
@bogdando: WDYT?

bogdando:
:+1:

PR #263, opened 2016-09-07
Fix example for auth parameter name in URI Query Parameters specification

skolodyazhnyy:
It seems like a typo, or may be outdated example.


michaelklishin:
Thank you very much! I will backport this to `live`.

PR #270, opened 2016-09-16
Remove links to empty dir listings

camelpunch:
Also, remove a trailing bar separator

[#128649497]


camelpunch:
Result is (currently) at http://prrabbitmq.cfapps.io/nightly-builds.html

PR #273, opened 2016-09-27
add note about paging using lots of memory

phunehehe:
basically writing down https://github.com/rabbitmq/rabbitmq-server/issues/965#issuecomment-247566352 for the next guy

### DIFF
--- a/site/memory.xml
+++ b/site/memory.xml
@@ -3,8 +3,8 @@
 Copyright (c) 2007-2016 Pivotal Software, Inc.
 
 All rights reserved. This program and the accompanying materials
-are made available under the terms of the under the Apache License, 
-Version 2.0 (the "License”); you may not use this file except in compliance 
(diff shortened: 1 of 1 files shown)


michaelklishin:
Sorry but this is a broad generalization. "Disc" is not a typo either.

phunehehe:
(You are so fast at closing stuff :D)
If "disc" is not a typo, perhaps the instance of "disk" should be converted?
I have nothing to say about the memory usage though, if you think it's an overgeneralization.

michaelklishin:
We can add a section on `vm_memory_high_watermark_paging_ratio` to make it more obvious. However, there is a number of guides that cover things around this area and it's not immediately clear where this belongs and what may need editing in those guides.

I've filed https://github.com/rabbitmq/rabbitmq-website/issues/274 for this.

PR #291, opened 2016-10-26
site/clustering.xml: Document that patch release may require a cluster shutdown

dumbbell:
[#132823995]


michaelklishin:
I think the two known exceptions should be listed as an, ahem, list, to make them more visible. We can document it as "known exceptions."

PR #306, opened 2016-11-28
Updated how.xml

packtpartner:
Added a new book link


michaelklishin:
Sorry but we won't be adding new content to that page. It badly need rewriting from scratch.

Issue #307, opened 2016-11-30
Changelog is missing 3.5.8


professor:
I noticed that the change log does not list 3.5.8, which appears to be the preferred 3.5.x release due to a security fix. 

http://www.rabbitmq.com/changelog.html

https://github.com/rabbitmq/rabbitmq-server/releases/tag/rabbitmq_v3_5_8

michaelklishin:
It [is not missing](https://github.com/rabbitmq/rabbitmq-website/blob/live/site/changelog.xml#L45).

michaelklishin:
3.5.8 is the second release from the top on http://www.rabbitmq.com/changelog.html (the list is in reverse chronological order).

professor:
We were upgrading a rabbit 3.5.2 installation to 3.5.7 and did not see 3.5.8 since it is not listed near the 3.5.x releases. We expect releases to be sorted by release number then date. 

Sorting by date, not release allowed us to miss an important security fix.

michaelklishin:
@professor I'm sorry but we can't assume it works that way for everybody. We announce releases on rabbitmq-users and Twitter (in fact, multiple times per release on Twitter). I suggest that you sign up for one of those.

We will not be changing the order based on one data point.

michaelklishin:
GitHub releases are also sorted by date (by tag date, to be precise), not version order.

PR #308, opened 2016-11-30
Adds 3.5.8 to change log

professor:
We were upgrading a rabbit 3.5.2 installation to 3.5.7 and did not see 3.5.8 since it is not listed near the 3.5.x releases. We expect releases to be sorted by release number then date. 

Fixes https://github.com/rabbitmq/rabbitmq-website/issues/307

### DIFF
--- a/site/changelog.xml
+++ b/site/changelog.xml
@@ -42,17 +42,6 @@ limitations under the License.
       </tr>
 
       <tr>
-        <td class="centre">3.5.8</td>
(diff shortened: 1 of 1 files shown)


michaelklishin:
Sorry but I'm not willing to change the order based on one data point (complaint).

michaelklishin:
GitHub releases are also sorted by date (by tag date, to be precise), not version order.

PR #313, opened 2016-12-16
New Relic <> RabbitMQ monitoring

jpatel3:
Added documentation link for monitoring RabbitMQ with New Relic


michaelklishin:
Thank you. That page will be replaced entirely in the future. A more suitable one is [Developer Tools](https://github.com/rabbitmq/rabbitmq-website/blob/master/site/devtools.xml). Also, please take a look at what branches are available in the repository (documented in README). `master` will be deployed in several months when `3.7.0` ships.

jpatel3:
Sure. Thanks for prompt response 👍

PR #322, opened 2017-01-05
Fix elixir RPC tutorial

Arkham:
* Fix typo in first snippet
* Adjust column width of code snippets so that it doesn't overflow


michaelklishin:
Thank you!

PR #323, opened 2017-01-09
Explain when an auto-delete queue will not be deleted

gerhard:
This was mentioned on the mailing list in the past, and the behaviour
definitely surprised me, capturing it here for posterity.


michaelklishin:
If this is intended to go live immediately it should have been submitted against `live`.

gerhard:
Can it be cherry-picked into `live`?

michaelklishin:
Already is.

PR #334, opened 2017-01-27
Document effect of partitions on management API/UI

### LINKED ISSUE #333: Document effects of network partitions on the management API/IU

kjnilsson:
Fixes #333

### DIFF
--- a/site/nettick.xml
+++ b/site/nettick.xml
@@ -91,5 +91,16 @@ node rabbit@cordelia down: net_tick_timeout
     ].</pre>
       </p>
     </doc:section>
+    <doc:section name="net-tick-frequency">
+      <doc:heading>Management API</doc:heading>
+      <p>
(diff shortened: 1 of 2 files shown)


site/nettick.xml
```diff
@@ -91,5 +91,16 @@ node rabbit@cordelia down: net_tick_timeout
     ].</pre>
       </p>
     </doc:section>
+    <doc:section name="net-tick-frequency">
+      <doc:heading>Management API</doc:heading>
```
michaelklishin: We don't really use the term "management API" much. Either "management UI" or "HTTP API". I suggest using the latter with a link to `/management.html`.

michaelklishin:
This should probably be based on `stable`. I'd also recommend adding a section to `/management.html` that links to the net tick guide.

kjnilsson:
Thanks. Updated.

PR #339, opened 2017-02-21
fix documentation to show that the default number of async threads is 64

komuW:
**What has changed:**

- updated documentation to reflect the fact that the default number of async threads is now 64 (as opposed to 30)

**Why**

- The default number of async threads was bumped to 64 from 30 in this pull requests;       

1. https://github.com/rabbitmq/rabbitmq-server/pull/212 
2. https://github.com/rabbitmq/rabbitmq-server/pull/704


michaelklishin:
Thank you.

komuW:
you are welcome

PR #340, opened 2017-03-03
Clarify memory alarm description

binarin:
The original "blocks all connections" is a bit misleading, as it implies that the connections are blocked indiscriminately.


michaelklishin:
@binarin I'm guessing this is meant for `live`? No need to re-submit, I will cherry pick, just making sure.

binarin:
Yes, for `live`.

michaelklishin:
Thank you, should already be live.

PR #341, opened 2017-03-07
add svg logo and replace png version

slang800:
Ever since I got a 4k screen, this has been bothering me. I found a full svg version of the RabbitMQ logo inside of one of the images on http://www.rabbitmq.com/blog/2010/10/19/exchange-to-exchange-bindings/#more-40, extracted it with inkscape, and optimized it with [svgo](https://github.com/svg/svgo). Also, now it lines up with the margin of the page, which is nice.

Before:

1488843573_after_aligned

After:

1488843587_after_aligned

Oh, and the logo looks a bit wider because the png version on the website header right now is resized from 277px wide to 253px. I removed the width attribute, so now it matches the copy on http://rabbitmq.com/trademark-guidelines.html


michaelklishin:
Thank you very much for you time! Unfortunately we cannot merge it as it loses the trademark bit (and yes, it is important to Pivotal's legal team).

The site is undergoing a redesign so we will see if an SVG version of the current logo makes sense after that's done.

slang800:
It's weird that your legal team would ask for that, given that almost no other companies put a TM icon in their logo, but I don't care enough to research it further.

If you reopen the PR I can add the TM bit... Pushing to a closed PR breaks stuff: https://github.com/isaacs/github/issues/361

dcarwin-pivotal:
All uses of the RabbitMQ design should be with the ® symbol. Thanks Sean.

michaelklishin:
@slang800 sorry but @dcarwin-pivotal suggests it should be the ® symbol, not ™.

slang800:
A copyright symbol would be an odd change, since your trademark-guideline page says the logo is a trademark. Are you sure @dcarwin-pivotal's message wasn't just mistyped?

With the new commit it looks like this:



If you want it changed I'll do that tomorrow.

michaelklishin:
Thank you. Our design has changed with #352. Please rebase and re-submit if you think this is still relevant.

slang800:
no need - the new site uses an SVG logo

PR #345, opened 2017-03-20
Using self descriptional constant

astahovn:


michaelklishin:
Thanks. Please re-submit this against the `live` branch. We'd also need to update the code in https://github.com/rabbitmq/rabbitmq-tutorials/.

PR #346, opened 2017-03-20
Using self descriptional constant

astahovn:


michaelklishin:
Thank you, it's already live.

PR #347, opened 2017-03-20
Missed other examples

astahovn:


astahovn:
Sorry, but i've missed the other two places with this code line

PR #349, opened 2017-03-24
minor update to tutorial-two-ruby.md

rsalgado:
Fix inconsistency in worker's code when sleeping.


michaelklishin:
Thanks. Please re-submit this against the `live` branch. It would be fantastic if you could also check the [runnable examples](https://github.com/rabbitmq/rabbitmq-tutorials).

PR #350, opened 2017-03-24
Minor update in Ruby Tutorial - Two

rsalgado:
Fixed an inconsistency in the worker code when sleeping to imitate some work.


michaelklishin:
Thank you!

PR #356, opened 2017-03-31
networking.xml: fix networking tunning link

chengweiv5:
Signed-off-by: Chengwei Yang <yangchengwei@qiyi.com>


michaelklishin:
I will cherry-pick to `live`. Thank you.

PR #357, opened 2017-03-31
networking.xml: fix tcp keepalive meaning

chengweiv5:
Refer to http://tldp.org/HOWTO/TCP-Keepalive-HOWTO/usingkeepalive.html

Signed-off-by: Chengwei Yang <yangchengwei@qiyi.com>


michaelklishin:
I will cherry-pick to `live`. Thank you.

PR #360, opened 2017-04-02
Update required php-amqplib version.

likeuntomurphy:
AMQPMessage::DELIVERY_MODE* constants used in the tutorials are not introduced in php-amqplib until 2.6.1.


michaelklishin:
Thank you so much! It's already live.

PR #363, opened 2017-04-04
Explain x-death entries are sorted chronologically

EdeMeijer:
I wasn't sure if `x-death` entries were sorted in any way, but found the discussion at https://github.com/rabbitmq/rabbitmq-server/pull/79 and it seems they are sorted by the last time a {queue, reason} happened. This might be useful information for the website.

### DIFF
--- a/site/dlx.xml
+++ b/site/dlx.xml
@@ -230,9 +230,10 @@ args.put("x-dead-letter-routing-key", "some-routing-key");</pre>
             expiring again in any queues it is routed to.</li>
         </ul>
 
-        In case <code>x-death</code> already contains an entry with
+        New entries are prepended to the beginning of the <code>x-death</code>
+        array. In case <code>x-death</code> already contains an entry with
         the same queue and dead lettering reason, its count field will be
-        incremented.
+        incremented and it will be moved to the beginning of the array.
       </p>
       <p>
         The <code>reason</code> is a name describing why the


site/dlx.xml
```diff
@@ -230,9 +230,10 @@ args.put("x-dead-letter-routing-key", "some-routing-key");</pre>
             expiring again in any queues it is routed to.</li>
         </ul>
 
-        In case <code>x-death</code> already contains an entry with
+        New entries are prepended to the beginning of the <code>x-death</code>
+        array. In case <code>x-death</code> already contains an entry with
         the same queue and dead lettering reason, its count field will be
-        incremented.
+        incremented and it will be moved to the beginning of the array.
```
michaelklishin: I think you mean "prepended", not "moved". Existing entries are not moved around.

michaelklishin:
Thank you. I think the wording could be improved somewhat. Also, please submit the same PR against the `live` branch if possible.

site/dlx.xml
EdeMeijer: Are they not? In https://github.com/rabbitmq/rabbitmq-server/issues/78 @simonmacmullen proposed this (emphasis mine):

> Look for a matching one, matching on both queue and reason
> If there, increment its count **and bring to the front of the list**
> If not, create one with count=1 and add to the front of the list

My Erlang isn't so great so I couldn't immediately spot if this was actually happening in the related commit, but assumed this is how it works now. It would be nice because it would make determining the last dead letter reason as easy as grabbing the first entry from x-death. If this is not the case, I will change the phrasing to make this clear.

So to be clear, in this scenario:
```
- rejected on queue foo at 00:00
- rejected on queue bar at 00:01
- rejected on queue foo at 00:02
- rejected on queue baz at 00:03
```
the order in x-death would be
```
- baz, 1x, last time at 00:03
- bar, 1x, last time at 00:01
- foo, 2x, last time at 00:02
```
?

michaelklishin: @EdeMeijer I'd need to check [the code](https://github.com/rabbitmq/rabbitmq-server/blob/stable/src/rabbit_dead_letter.erl) since I don't remember off the top of my head. Well, the existing entries are updated so it's a matter of wording.

The focus of https://github.com/rabbitmq/rabbitmq-server/issues/78 wasn't so much to enforce a particular ordering.

michaelklishin: Yes, looks like we do [move updated values](https://github.com/rabbitmq/rabbitmq-server/blob/stable/src/rabbit_dead_letter.erl#L138) to the beginning of the list. So perhaps your wording is clearer (and more correct) than mine 👍 .

michaelklishin:
I will cherry-pick to `live`.

site/dlx.xml
EdeMeijer: Good to know!

PR #370, opened 2017-04-17
Fix a typo in a PHP tutorial

Kolyunya:
The `routing_key` [seems to be](https://github.com/php-amqplib/php-amqplib/blob/cb0433a8d1ab2d7952ae02bc7433ad181a0f1446/PhpAmqpLib/Channel/AMQPChannel.php#L1098) the third parameter, not the second.


michaelklishin:
Will cherry-pick to `live`.

PR #377, opened 2017-05-01
elixir: use 0.2.1 now its released

dch:
refer to https://github.com/rabbitmq/rabbitmq-tutorials/pull/133 for matching commit


michaelklishin:
Thank you!

PR #378, opened 2017-05-08
Clarify queue synchronisation concepts

gerhard:
In a recent discussion it became obvious that some concepts are worth reiterating in different contexts, in particular **Configuring Synchronisation**.

Made some formatting improvements while at it, just enough to make it easier to follow.


gerhard:
Will need merging into master when we're happy with the changes

gerhard:
💯

PR #379, opened 2017-05-11
Deploy with interactive defaults

gerhard:
I had to deploy recently and would have liked sensible defaults with prompts. What do you think?

### DIFF
--- a/deploy
+++ b/deploy
@@ -1,20 +1,60 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 set -e
 
-remote_host=$1
-mode=$2
-version=$3
+REMOTE_HOST="$1"
+DEFAULT_REMOTE_HOST="www.rabbitmq.com"
 
-if [ x"$remote_host" = x -o x"$mode" = x -o \( x"${version}" = x -a x"$mode" = xprevious \) ] ; then
-    echo 'Usage: deploy <host to deploy to> [www|next|previous vA_B_C]'
-    exit 1
-fi
+VERSION="$2"
+DEFAULT_VERSION=www
+VALID_VERSIONS=( previous www next )
 
-tmp_dir=/tmp/to_deploy_${mode}
+GIT_SHA="$3"
 
-set -x
+main() {
+  debug
+  set_remote_host
+  set_version
+  deploy
+}
 
-ssh $remote_host mkdir -p $tmp_dir
-rsync -qcrz --delete --perms --chmod=a+rwX --exclude=site/man code conf site wordpress-theme $remote_host:$tmp_dir
-ssh -t $remote_host sudo "$tmp_dir/code/deploy-remote ${mode} ${version}"
+debug() {
+  [ -z "$DEBUG" ] || set -x
+}
+
+set_remote_host() {
+  if [ -z "$REMOTE_HOST" ]
+  then
+    echo
+    read -rp "Which user & host do you want to deploy to? (press enter for $DEFAULT_REMOTE_HOST): " remote_host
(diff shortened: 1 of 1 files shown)


michaelklishin:
I don't see much need for interactive inputs as we have only 2 required parameters. This PR produces no deployment command output for me either interactively or via `./deploy {username}@next.rabbitmq.com next`.

michaelklishin:
That said, if positional arguments continue working just like today, why not. But currently I get no output from any `ssh` commands, so something is wrong.

Also, this PR was probably meant to be submitted against `live`.

gerhard:
This PR does not change the current behaviour when using positional arguments, so it's fully backwards compatible.

The interactive nature and sensible defaults kick in when running just `./deploy`

gerhard:
If I run it in debug mode (`DEBUG=1`) I get the following:

```sh
+ set -e
+ REMOTE_HOST=gerhard@www.rabbitmq.com
+ VERSION=www
+ VERSIONS=(previous www next)
+ GIT_SHA=
+ '[' -n '' ']'
+ echo

+ read -rp 'Which user & host do you want to deploy to? (press enter for gerhard@www.rabbitmq.com): ' remote_host
Which user & host do you want to deploy to? (press enter for gerhard@www.rabbitmq.com):
+ REMOTE_HOST=gerhard@www.rabbitmq.com
+ '[' -n '' ']'
+ echo

+ read -rp 'Which website version is this? (press enter for www or choose one of: previous www next): ' version
Which website version is this? (press enter for www or choose one of: previous www next):
+ VERSION=www
+ TMP_DIR=/tmp/to_deploy_www
+ ssh gerhard@www.rabbitmq.com mkdir -p /tmp/to_deploy_www
+ rsync -qcrz --delete --perms --chmod=a+rwX --exclude=site/man code conf site wordpress-theme gerhard@www.rabbitmq.com:/tmp/to_deploy_www
+ ssh -t gerhard@www.rabbitmq.com sudo '/tmp/to_deploy_www/code/deploy-remote www '
[sudo] password for gerhard:
Feeds will be mirrored by the user rabbitmq
######.#.#####.###########.##.#..#######.#####.##.###.#..##.#..#.##.##.####.##.#.########.#######...#####.#..###.##.###########.#.#########.######.###.#.##########.###.....######################################.#############.########.######....................................................................................................................................######........................................................................................................................................................#
Site www.rabbitmq.com already enabled
[ ok ] Reloading web server config: apache2.
Connection to www.rabbitmq.com closed.
```

deploy
```diff
@@ -1,20 +1,36 @@
-#!/bin/sh
+#!/bin/bash
```
dumbbell: This is not portable :) Please use `/usr/bin/env bash` if you really want to use Bash.

deploy
```diff
@@ -1,20 +1,36 @@
...
+
+[ -z "$DEBUG" ] || set -x
 
 set -e
 
-remote_host=$1
-mode=$2
-version=$3
+REMOTE_HOST="$USER@www.rabbitmq.com"
```
dumbbell: There is no need to prepend `$USER@`. ssh(1) will do that itself and you could also set the user in your ssh_config(5).

dumbbell:
@michaelklishin has a point: the script in `live` calls `set -x` before starting the actual deploy. The new script should keep this behavior, even if you set it earlier with `DEBUG=1`. Otherwise, it asks for a password but we don't have any context.

gerhard:
👀

gerhard:
@michaelklishin 👀

PR #380, opened 2017-05-16
add more rabbitMQ saas companies

komuW:
## What has changed.
- Added links to two more rabbitMQ as a service companies that have emerged in the past few years.

## Why has it changed.
- The rabbitMQ website already contains links to one rabbitMQ as a service company, I think it makes sense for the website to reflect all the rabbitMQ service companies that are available today.

### DIFF
--- a/site/download.xml
+++ b/site/download.xml
@@ -99,6 +99,8 @@ limitations under the License.
         <li><a href="https://github.com/pivotal-cf/cf-rabbitmq-release">Cloud Foundry</a></li>
         <li><a href="http://docs.pivotal.io/rabbitmq-cf/index.html">Pivotal Cloud Foundry</a></li>
         <li><a href="https://www.cloudamqp.com">CloudAMQP</a>: RabbitMQ-as-a-Service available in multiple clouds</li>
+        <li><a href="https://compose.com/rabbitmq">Compose</a>: Hosted RabbitMQ</li>
+        <li><a href="https://www.rabbitmqhosting.com">RabbitmqHosting</a>: RabbitMQ as a service</li>
       </ul>
 
       <h3>Chef, Puppet, Docker</h3>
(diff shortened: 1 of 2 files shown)


gerhard:
@komuW Compose is a good addition, not so sure about the other company. We don't want an exhaustive list, I'm happy if we just add Compose. @michaelklishin what do you think?

michaelklishin:
We discussed this with @dcarwin-pivotal and the conclusion is this: Compose seems to be a reasonably mature option worth adding. RabbitmqHosting is very new and unproven. It also has a name that may lead some to believe that it is a Pivotal-backed project (which is not the case).

I will add Compose manually.

komuW:
@gerhard I don't see what would be the problem of the official rabbitMQ website providing an exhaustive list of rabbitMQ SaaS companies. Unless there are some SaaS companies that are endorsed by RMQ/pivotal. But if RMQ/pivotal are neutral, then I feel like -IMHO- the RMQ website should have a page with a list of as many RMQ Saas companies as can be found and then the readers are left to figure out which of those they want to go with. 

@michaelklishin I'm no lawyer so I don't know how the name RabbitmqHosting would lead anyone to think that it is affiliated with pivotal, unless they(RabbitmqHosting) are appropriating some of pivotal's copyright materials/names. 
I however do agree that RabbitmqHosting is new and unproven, so for that reason alone I would agree with you that maybe it shouldn't be included. 

I think in conclusion, This brings to the fore the question:
1. Should the rabbitMQ website have a list of RMQ SaaS companies? And if yes, what is the criteria for admission? Should that criteria be laid out in writing or should it be left to the discretion of people during code review?

michaelklishin:
@komuW the criteria for admission is this: our team sees enough traction around the service.

I'm not a lawyer either but pretty sure "RabbitMQ" is a trademark registered by Pivotal.

komuW:
@michaelklishin my bad. I hadn't looked at the RabbitMQ trademark page(I wasn't even aware there was one).  My mistake, not yours.

So having gone through that page(https://www.rabbitmq.com/trademark-guidelines.html), I would have to agree with you. 
I withdraw most of my earlier points, but I still stand with this one point:
`I feel like -IMHO- the RabbitMQ website should have a page with a list of as many RabbitMQ Saas companies as can be found and then the readers are left to figure out which of those they want to go with.`

michaelklishin:
@komuW there's already a page that can provide pretty much all the options: Google search.

PR #381, opened 2017-05-19
Improve production memory guidelines

gerhard:
This is just a first pass, it needs a few more refinements and a few
more opinions. It's already better than what we had, but not there yet.

Let's start discussing: @michaelklishin @dumbbell @Gsantomaggio

### DIFF
--- a/site/production-checklist.xml
+++ b/site/production-checklist.xml
@@ -101,17 +101,16 @@ limitations under the License.
         <doc:heading>Memory</doc:heading>
         <p>
           By default, RabbitMQ will not accept any new messages when it detects
-          that it's using more than 40% of the available system memory
-          (<code>vm_memory_high_watermark, 0.4</code>). This is a safe default
-          and care should be taken when modifying this value, even if the host
-          is a dedicated RabbitMQ node. <a href="/memory.html">Learn more about
-            RabbitMQ &amp; system memory</a>.
+          that it's using more than 40% of the available system memory:
+          <code>{vm_memory_high_watermark, 0.4}</code>. This is a safe default
+          and care should be taken when modifying this value, even when the host
+          is a dedicated RabbitMQ node.
         </p>
         <p>
           The OS and file system use system memory to speed up operations for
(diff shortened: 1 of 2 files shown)


site/production-checklist.xml
```diff
@@ -100,38 +100,38 @@ limitations under the License.
       <doc:subsection name="resource-limits-ram">
         <doc:heading>Memory</doc:heading>
         <p>
-          By default, RabbitMQ will use up to 40% of available RAM. For nodes
-          that are dedicated to run RabbitMQ, it is often reasonable to raise
-          the limit. However, care should be taken as the OS and file system caches
-          need RAM to operate as well. Failing to do so will have a severe throughput drop due
-          to OS swapping, or even result in the RabbitMQ process terminated by the OS.
+          By default, RabbitMQ will reject any new messages when it detects
```
Gsantomaggio: I am not totally sure about `will reject` 
I'd say `won't accept` ( or similar) since reject could be confused with ` basic.reject`

gerhard:
Having reviewed this with a fresh perspective, this improvement feels sufficient.

@michaelklishin / @dumbbell please merge if happy with the changes.

site/production-checklist.xml
```diff
@@ -100,38 +100,40 @@ limitations under the License.
...
             <li>
-              75% of the configured RAM limit when
-              the limit is between 8 and 16 GB of RAM
-            </li>
-            <li>
-              80% of the configured RAM limit when
-              the limit is above 16 GB of RAM
+              Values above <code>0.7</code> are not recommended. The OS and
+              file system must be left with at least 30% of all available system
+              memory, otherwise performance will degrade severely.
```
dumbbell: I would say "otherwise performance may degrade severely" because we can't know.

Also, I would conclude that paragraph with something like:
> As always, proper benchmarking and measurements are required to find the best setting for your workload.

gerhard:
@dumbbell great suggestions, thanks! Good to merge?

PR #387, opened 2017-05-30
Document JNDI support in WildFly (JMS client)

acogoluegnes:
Part of rabbitmq/rabbitmq-jms-client#21


michaelklishin:
We can't merge this into `live` without releasing JMS client `1.7.0` first.

michaelklishin:
We have an RC out => merging.

PR #398, opened 2017-06-27
Document snapshots, replacing nightlies

dumbbell:
Nightlies are not produced for months. They are replaced by snapshots published by our CI everytime testing is successful.

In addition to what was said about nightlies, explain how to setup Debian and Yum repositories.

Currently, snapshots are made available on Bintray repositories only.

[#128649497]


dumbbell:
A copy of the website with this patch was published here (until another pull request overwrites it):
https://prrabbitmq.cfapps.io/snapshots.html

PR #402, opened 2017-07-11
Correct the description example URI

russelldb:
And correct the tabular description of `connection_timeout`. It appears from testing and https://www.rabbitmq.com/releases/rabbitmq-erlang-client/v3.6.10/doc/amqp_connection.html that `connection_timeout` is in fact a millisecond, not second, value.


michaelklishin:
Thank you!

I will cherry-pick to `live`.

PR #405, opened 2017-07-24
Mention how to reference System.Threading.

kjnilsson:
To clarify that it is required without having to inspect the full sample
at the bottom of the file.

Fixes: https://github.com/rabbitmq/rabbitmq-tutorials/issues/136

Not sure if this needs to go to master or stable.


michaelklishin:
It needs to go to `live`, I will cherry-pick.

PR #411, opened 2017-08-04
Fixed a sentence regarding autoack in the tutorials. 

chaitanyagurrapu:
I am still going through the tutorials, so I could be wrong, but I think this is what you mean?


michaelklishin:
`autoAck` does not turn manual acknowledgements on, it turns them off.

chaitanyagurrapu:
Ah. That makes sense. I kept thinking "but how are we turning the acknowledgements **off** by setting the 'autoack' flag to true?". Sent another PR with slight modifications to clarify where we are referring to manual acknowledgement and where we are referring to auto acknowledgement.

PR #414, opened 2017-08-04
Updating based on conversation in pull request

chaitanyagurrapu:


chaitanyagurrapu:
Will try to find the cycles to make the changes to the other language tutorials.

Not sure what you mean by point 1 though. Should I just add the above link at the end of the paragraph?

I will read the link above and finish the tutorials and maybe it will all make sense.

michaelklishin:
I mean that "automatic mode" could be "[automatic mode](http://www.rabbitmq.com/confirms.html)" :) thank you!

michaelklishin:
I will apply this manually to all tutorials. Thank you.

michaelklishin:
Actually, this cannot be applied to all tutorials because the default is not the same in all client libraries. I will revisit them one by one.

PR #430, opened 2017-09-15
Updated tutorial-six-java.md with as per changes in RPCServer.java

2sbsbsb:
a. Channel is final
b. Replace sleep() with wait()/notify()


michaelklishin:
@2sbsbsb thank you. I will cherry-pick to `live` so it can go live immediately.

PR #431, opened 2017-09-15
Update client code in the article

darrylsk:
Update the article to reflect an update to the csharp client code that replaced the usage of an obsolete class with a current one.


michaelklishin:
Please rebase and re-submit this against the `live` branch since it will have to go live immediately with the rabbitmq-tutorials changes.

PR #434, opened 2017-09-19
Add  msg_store_credit_disc_bound  & msg_store_credit_disc_bound to https://www.rabbitmq.com/configure.html

gerhard:
Should be merged with rabbitmq/rabbitmq-server#1366


gerhard:
There are other properties which I wanted to initially add, but rabbitmq/rabbitmq-server#1366 is already bigger than anticipated, don't want to add anymore scope to it.

This is now ready to review and merge on its own.

PR #448, opened 2017-10-16
Fix second python tutorial

sazarkin:
New queue name is added in code example too early

### DIFF
--- a/site/tutorials/tutorial-two-python.md
+++ b/site/tutorials/tutorial-two-python.md
@@ -83,11 +83,8 @@ import sys
 
 message = ' '.join(sys.argv[1:]) or "Hello World!"
 channel.basic_publish(exchange='',
-                      routing_key='task_queue',
-                      body=message,
-                      properties=pika.BasicProperties(
-                         delivery_mode = 2, # make message persistent
-                      ))
+                      routing_key='hello',
+                      body=message)
 print(" [x] Sent %r" % message)
 </pre>
 


pivotal-issuemaster:
@sazarkin Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=448)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=448) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sazarkin Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=448)!

michaelklishin:
Thank you but I don't think I agree. The paragraph right above this snippet mentions a new program. In fact, I believe we had the proposed version, intentionally or not, at some point and it was seen as an issue by others.

sazarkin:
Ok, it is up to you.
But there is issue, in tutorial text or code. 
Old program and this code snippet does not sync, and it looks like it was copy pasted from second half of this tutorial.
Cheers.

uolot:
I came here to create the very same PR - this bit of example code doesn't match what the surrounding text suggests. 

Changing the `routing_key` is explicitly mentioned in the [Message durability](https://github.com/rabbitmq/rabbitmq-website/blob/live/site/tutorials/tutorial-two-python.md#message-durability) part of this tutorial:

> RabbitMQ doesn't allow you to redefine an existing queue with different parameters and will return an error to any program that tries to do that. But there is a quick workaround - let's declare a queue with different name, for example task_queue

Same for `delivery_mode`, a few lines down from the above:

> At that point we're sure that the task_queue queue won't be lost even if RabbitMQ restarts. Now we need to mark our messages as persistent
> * by supplying a delivery_mode property with a value 2.

Yet, these changes are introduced in the snippet that this PR tries to update. Doing the tutorial for the first time it's confusing and the PR should be accepted.

michaelklishin:
Sentences prior to the modified example talk about modifying a tutorial 1 version which does not have additional message properties. Fair enough.

michaelklishin:
Cherry-picked to `live` and it is already, well, live. Thank you.

PR #466, opened 2017-12-13
Replace broken link about use as a job queue

schwarz:
The author of "Generating Thousands of PDFs on EC2 with Ruby" reposted it on his blog.


michaelklishin:
Thank you. Boy does that page need a major overhaul. Links from 2009 😱.

PR #475, opened 2018-01-15
Tutorial fix for golang

magicalbanana:
👨 🐻 🐷 

The tutorial is incorrect because Ack is turned on by default and
when acking a new mesage, `true` must be passed, not `false`.

### DIFF
--- a/site/tutorials/tutorial-two-go.md
+++ b/site/tutorials/tutorial-two-go.md
@@ -228,8 +228,7 @@ There aren't any message timeouts; RabbitMQ will redeliver the message when
 the consumer dies. It's fine even if processing a message takes a very, very
 long time.
 
-Message acknowledgments are turned on by default.
-It's time to turn them off using the `false,  // auto-ack` option and send a proper acknowledgment
+To turn off auto acknowledgment, pass `false, // auto-ack` and send a proper acknowledgment
 from the worker `d.Ack(true)`, once we're done with a task.
 
 <pre class="sourcecode go">
--- a/site/tutorials/tutorial-two-go.md
+++ b/site/tutorials/tutorial-two-go.md
@@ -228,9 +228,9 @@ There aren't any message timeouts; RabbitMQ will redeliver the message when
 the consumer dies. It's fine even if processing a message takes a very, very
 long time.
 
-Message acknowledgments are turned off by default.
-It's time to turn them on using the `false,  // auto-ack` option and send a proper acknowledgment
-from the worker `d.Ack(false)`, once we're done with a task.
+Message acknowledgments are turned on by default.
+It's time to turn them off using the `false,  // auto-ack` option and send a proper acknowledgment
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@magicalbanana Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=475)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=475) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@magicalbanana Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=475)!

michaelklishin:
Thank you for your time. I'm not sure I'm convinced by this change.

First, let's clarify the terminology. I'll assume "Ack" means "[manual acknowledgements](http://www.rabbitmq.com/confirms.html)". If that's the case, the tutorial code [already enables them by passing a false here](https://github.com/rabbitmq/rabbitmq-tutorials/blob/master/go/worker.go#L46).

The second change calls `Ack` on [a delivery](https://github.com/streadway/amqp/blob/master/delivery.go#L123), which *assumes* that manual acknowledgements are used and simply decides whether a single delivery is acknowledged or multiple ones. Both options are valid, in particular for the sake of this tutorial which does not really try to explain the multi-ack scenario.

So can we take a step back and describe what problem we are solving here and how it can be verified?

michaelklishin:
Unlike in many other clients, with the Go one the user must specify every single parameter that will be set on a `basic.consume` method on the wire. Therefore the default is what the user decides it to be. I suspect the copy is a leftover from the original tutorial this one is based on.

So while the copy does need a bit of work — and I'm happy to correct them — the idea that this client has a default mode for message acknowledgements is not really correct. Acknowledging one delivery at a time makes more sense to me (although both can work perfectly fine) because the tutorial just tries to demonstrate the concept of manual acknowledgements, not e.g. how to increase throughput for a fast consumer or that such a feature exists in the protocol.

magicalbanana:
@micade ah, gotcha. The only thing I wanted to change on the tutorial was when passing `fasle` to `Ack` it stated that it acknowledged the message, when if fact, that would mean it didn't.

magicalbanana:
And you're right, `Ack` is not turned on by default because if you didn't pass anything, then it would mean that the zero value for that attribute would be false.

magicalbanana:
And feel free to correct it, we can close this PR if you wish.

gerhard:
👍

michaelklishin:
We will correct the copy and give @magicalbanana credit directly in `live`. Thanks for bringing this up.

I don't think that tutorial 2 should acknowledge multiple deliveries at a time.

PR #487, opened 2018-01-25
XSLT: Make all headings for named sections into links

hairyhum:
Currently there are anchors, but no links to get the anchors from.
Make it easier to link to specific sections by turning section
headings into links.

### DIFF
--- a/site/access-control.xml
+++ b/site/access-control.xml
@@ -112,7 +112,7 @@ loopback_users = none
 </pre>
       </doc:section>
 
-      <doc:section name="access-control">
+      <doc:section name="permissions">
         <doc:heading>How Permissions Work</doc:heading>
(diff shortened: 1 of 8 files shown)


michaelklishin:
I dared to change the anchor links to be different from other links:

 * They are non-underlined by default, underlined on hover
 * They are in [dark grey](http://coloreminder.com/a9a9a9) instead of bright orange.

Let's see how this goes and if there are any complaints. All page headers becoming anchors is definitely handy.

hairyhum:
On `.md` pages header links don't have `anchor` class, so they will be different from `.xml` pages. Should we change them too?

michaelklishin:
Yes, please. That class was meant to be used for all anchors.

PR #530, opened 2018-05-12
Priority Queues guide improvements

michaelklishin:
Documentation improvements to go with rabbitmq/rabbitmq-server#1590.

### DIFF
--- a/site/priority.xml
+++ b/site/priority.xml
@@ -31,7 +31,7 @@ limitations under the License.
       <p>
         RabbitMQ has priority queue implementation in the core as of version <code>3.5.0</code>.
         Any queue can be turned into a priority one using <a href="/queues.html#optional-arguments">optional arguments</a>.
-        The implementation supports a limited number of priorities.
+        The implementation supports a limited number of priorities: 255. Values between 1 and 10 are recommended.
       </p>
     </doc:section>
 
--- a/site/priority.xml
+++ b/site/priority.xml
@@ -48,7 +48,7 @@ limitations under the License.
         Here's an example that will make all queues whose name begin with "priority." to have
         up to 5 priorities:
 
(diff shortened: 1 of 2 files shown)


site/priority.xml
```diff
@@ -27,127 +27,172 @@ limitations under the License.
...
-      indicating the maximum priority the queue should support. For example,
-      using the Java client:
-    </p>
-
-    <pre class="sourcecode java">
+    <doc:section name="intro">
+      <p>
+        RabbitMQ has priority queue implementation in the core as of version <code>3.5.0</code>.
+        Any queue can be turned into a priority one using <a href="/queues.html#optional-arguments">optional arguments</a>.
+        The implementation supports a limited number of priorities.
```
acogoluegnes: Maybe we should mention the range for priorities, as the exact values show up only in the "Using Client-provided Optional Arguments" section, not in the policy one.

site/priority.xml
```diff
@@ -27,127 +27,172 @@ limitations under the License.
...
+        Publishers then can publish prioritised messages using the
+        <code>priority</code> field of
+        <code>basic.properties</code>. Larger numbers indicate higher
+        priority.
+      </p>
+
+      <p>
+        Because the on-disk format for a priority queue is different,
+        <b>priority queues can only be defined by arguments</b>, not
+        policies. Queues can never change the number of priorities they
```
acogoluegnes: "not policies" is misleading as this is the first way we use to define priorities.

michaelklishin: That comment explains why policies are probably not a good idea in this case. This must be partially undone then 🤦‍♂️.

michaelklishin:
The docs were updated to match https://github.com/rabbitmq/rabbitmq-server/commit/08168de6b0c3c4d7f6429f2c3d098e5080ff7b1a.

PR #537, opened 2018-05-28
PHP Tutorial pages conform to PSR-2 standard

brianshortnh:
The code in the PHP tutorials was sometimes a bit messy.
With these changes the code conforms to standards outlined here:
https://www.php-fig.org/psr/psr-2/


michaelklishin:
Thank you!

PR #543, opened 2018-06-14
Update SpringOne Platform promo banner, add git commit script

jacksoncvm:


jacksoncvm:
Ok, I'll take that out.

PR #552, opened 2018-07-02
Update Supported Erlang versions guide to mention OTP 21

michaelklishin:
Note that I now list 19.3.6.4 as the minimum required. Even though nodes will still run on 19.3 we will ask any pre-19.3.6.4 users to upgrade first since [ERL-430](https://bugs.erlang.org/browse/ERL-430) and [ERL-448](https://bugs.erlang.org/browse/ERL-448) are catastrophic and running into them on affected versions is a matter of time. So this is a strong recommendation stated as a requirement, not a breaking change.

References rabbitmq/rabbitmq-server#1616.

[#157964874]


gerhard:
Enforcing Erlang/OTP 19.3.6.4 as the minimum supported version is the right call. We've been reasonable and only kept mentioning it for 6 months, it was time to pull the trigger. This is good!

PR #558, opened 2018-07-12
Flag is `-detached` not `--detached`

blgm:
The flag `--detached` is ignored by `rabbitmq-server`.  The documentation should consistently use the correct `-detached` flag.  All other occurrences that I could find (including the man page) are correct.


michaelklishin:
Thank you!

PR #584, opened 2018-09-10
Initial Quorum Queues doc guide

kjnilsson:


site/quorum-queue.md
```diff
@@ -0,0 +1,216 @@
...
+-->
+
+# Quorum Queue NOSYNTAX
+
+## Overview
+
+The quorum queue is a new (since 3.8) queue type for RabbitMQ implementing a persistent,
+replicated fifo queue using [the Raft consensus algorithm](https://raft.github.io/).
+
+The quorum queue type is an alternative to persistent HA queues for use cases
```
acogoluegnes: Should we use "HA queues" or "mirrored queues"? I don't know the official name, but I'm more used to "mirrored queues" personally.

site/quorum-queue.md
```diff
@@ -0,0 +1,216 @@
...
+RabbitMQ queues.
+
+It can also be created through the management UI using the queue type drop down
+menu.
+
+#### [Controlling the quorum queue size](#controlling-size)
+
+By default the quorum queue will launch a member node on every configured RabbitMQ
+node. This may not be desirable for very large RabbitMQ clusters or for cluster with
+an uneven number of RabbitMQ nodes (see below).
```
acogoluegnes: Link to the section?

site/quorum-queue.md
```diff
@@ -0,0 +1,216 @@
...
+#### [Changing quorum queue members](#changing-members)
+
+Once declared the RabbitMQ nodes a quorum queue resides on won't change even if the
+members of the RabbitMQ cluster change. To re-balance after a RabbitMQ cluster change
+quorum queues will have to be manually adjusted using the `rabbitmq-queues`
+command line tool.
+
+Two commands are provided:
+
+`rabbitmq-queues add_member <queue-name> <node>`
```
acogoluegnes: No parameter for virtual host?

site/quorum-queue.md
```diff
@@ -0,0 +1,216 @@
...
+
+Performance tails off quite a bit for quorum queue node sizes larger than 5.
+We do not recommend running quorum queues on more than 7 RabbitMQ nodes which is
+the default value of the `x-quorum-queue-size` queue argument.
+
+#### [Data safety](#data-safety)
+
+Quorum queues are designed to provide data safety under network partition and
+failure scenarios. A message that was successfully confirmed back to the publisher
+using the publisher confirms feature should not be lost as long as at
```
acogoluegnes: Link to publisher confirms page?

site/quorum-queue.md
```diff
@@ -0,0 +1,216 @@
...
+| Lazy | yes | no |
+| Priority | yes | no |
+| Dead letter queues | yes | yes |
+| Adheres to policies | yes | no |
+| Reacts to memory alarm | yes | no |
+
+
+## [Primary use case](#use-case)
+
+Quorum queues are _not_ designed to be used everywhere. The intended use are
```
acogoluegnes: "uses are"?

site/quorum-queue.md
kjnilsson: Good question. If you feel "mirrored queues" is the more common term I can use that instead. Or perahps I could write "Highty Available (Mirrored) queues" just like the doc page: https://www.rabbitmq.com/ha.html :)

acogoluegnes: Yes, I think "highly available (mirrored) queues" with a link to the section for the first reference would work. Then simply "HA queues" is good. Maybe stick with the link to the section, at least for the latest references.

site/quorum-queue.md
```diff
@@ -0,0 +1,266 @@
...
+| 7          | 3        |
+
+As the table above shows RabbitMQ clusters with fewer than 3 nodes do not
+benefit fully from the quorum queue guarantees. RabbitMQ clusters with an even
+number of RabbitMQ nodes do not benefit from having quorum queue members spread
+over all nodes. For these systems the quorum queue size should be constrained to a
+smaller uneven number of nodes.
+
+Performance tails off quite a bit for quorum queue node sizes larger than 5.
+We do not recommend running quorum queues on more than 7 RabbitMQ nodes which is
```
dcorbacho: Defaults to `5`

site/quorum-queue.md
```diff
@@ -0,0 +1,267 @@
...
+To declare a quorum queue set the `x-queue-type` queue argument to `quorum`
+(the default is `classic`). This will declare a quorum queue with a member on
+every currently configured RabbitMQ node.
+After that the queue can be bound to any exchange just as any other
+RabbitMQ queues.
+
+It can also be created through the management UI using the queue type drop down
+menu.
+
+#### [Controlling the quorum queue size](#controlling-size)
```
hairyhum: The `quorum-queue-size` name may be confused with a queue depth. Maybe it can be renamed to reflect the fact that this is a limit of nodes members of the queue cluster?

site/quorum-queue.md
```diff
@@ -0,0 +1,267 @@
...
+
+By default the quorum queue will launch a member node on every configured RabbitMQ
+node. This may not be desirable for very large RabbitMQ clusters or for cluster with
+an uneven number of RabbitMQ nodes.
+To control the number of quorum queue members set the `x-quorum-queue-size`
+queue argument when declaring the queue. The queue size argument provided should
```
hairyhum: Clarify what happens when it's greater than a rabbitmq cluster size.

Say it's configured to 7, but the cluster has 3 nodes, will the queue be successfully declared and start?
If later rabbitmq cluster is resized to 5 or 7 nodes, will new nodes get queue followers?
If rabbitmq cluster is resized down, does the queue still have quorum?

site/quorum-queue.md
```diff
@@ -0,0 +1,267 @@
...
+#### [Changing quorum queue members](#changing-members)
+
+Once declared the RabbitMQ nodes a quorum queue resides on won't change even if the
+members of the RabbitMQ cluster change. To re-balance after a RabbitMQ cluster change
+quorum queues will have to be manually adjusted using the `rabbitmq-queues`
+command line tool.
+
+Two commands are provided:
+
+`rabbitmq-queues add_member [-p <vhost>] <queue-name> <node>`
```
hairyhum: Clarify that `add_member` is not limited by the size configuration.

site/quorum-queue.md
```diff
@@ -0,0 +1,267 @@
...
+
+Quorum queues should be able to tolerate a minority of queue members becoming unavailable
+with no or little affect on availability. (NB: RabbitMQ may restart itself during
+recovery and perform various actions that make this harder to ensure but in
+principle this should be true).
+
+If a quorum of nodes cannot be recovered (say if 2 out of 3 RabbitMQ nodes are
+permanently lost) the queue is permanently unavailable and
+will need to be force deleted and recreated.
+
```
hairyhum: This new queue type will be used to store data, hence the queue depth can be high.
There should be a section on resource (disk and memory) usage and a warning that you can't have infinite queue because there is no flushing to disk.

site/quorum-queue.md
```diff
@@ -0,0 +1,267 @@
...
+a leader change
+which means that availability should be resumed very shortly after the failure
+event even when the queue has a significant backlog.
+
+## [Performance](#performance)
+
+Quorum queues are designed to trade latency for throughput and have been tested
+and compared against HA queues in 3, 5 and 7 node configurations at several
+message sizes. In all scenarios quorum queues have been observed to be have
+higher or equal throughput to HA queues. When message sizes are small they can
```
hairyhum: This is a bold statement. Some users can see that as a promise. Is there some benchmark to reproduce with specific conditions description?

site/quorum-queue.md
```diff
@@ -0,0 +1,267 @@
...
+Currently you have to use the [advanced](configure.html#advanced-config-file) RabbitMQ config for this.
+
+The `ra` application (which is the Raft library that quorum
+queues use) has it's own configuration.
+See [https://github.com/rabbitmq/ra#configuration](https://github.com/rabbitmq/ra#configuration)
+for details.
+
+The `rabbit` application has a couple of quorum queue related configuration items:
+
+* `quorum_cluster_size`
```
hairyhum: What happens with already declared queues if this setting is changed?

site/quorum-queue.md
kjnilsson: "have been observed" does not equal any kind of promise.

hairyhum: I worry about "In all scenarios" statement. Can we still produce specific scenarios description?

kjnilsson: "all tested scenarios"? I don't want to put specifics in a generic doc that will needs to be relevant as the feature evolves. We can also include a section on how the users will need to compare their own specific scenarios to properly evaluate performance.

michaelklishin: "In all tested scenarios" sounds reasonable. We can also say "most scenarios", too.

site/quorum-queue.md
michaelklishin: @kjnilsson and agreed on `x-quorum-initial-group-size` for the name, with `x-quorum-` being a common prefix for quorum queue arguments.

PR #598, opened 2018-10-05
DO NOT MERGE YET Document RPC support with JMS client

acogoluegnes:
Should be merged when JMS Client 1.11.0 is released.

### DIFF
--- a/site/jms-client.md
+++ b/site/jms-client.md
@@ -367,6 +367,151 @@ may have configuration files or command-line options.
 Refer to the documentation for the target logging framework
 for configuration details.
 
+## <a id="rpc-support" class="anchor" href="#rpc-support">Support for Request/Reply (a.k.a. RPC)</a>
+
+It is possible to use JMS for synchronous request/reply use cases.
+This pattern is commonly known as *Remote Procedure Call* or *RPC*.
+
+### With JMS API
+
+An RPC client can be implemented in pure JMS like the following:
+
+<pre class="sourcecode java">
+Message request = ... // create the request message
+// set up reply-to queue and start listening on it
+Destination replyQueue = session.createTemporaryQueue();
+message.setJMSReplyTo(replyQueue);
+MessageConsumer responseConsumer = session.createConsumer(replyQueue);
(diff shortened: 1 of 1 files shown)


site/jms-client.md
```diff
@@ -367,6 +367,164 @@ may have configuration files or command-line options.
...
+This pattern is commonly known as *Remote Procedure Call* or *RPC*.
+
+### With JMS API
+
+An RPC client can be implemented in pure JMS like the following:
+
+<pre class="sourcecode java">
+Message request = ... // create the request message
+// set up reply-to queue and start listening on it
+Destination replyQueue = session.createTemporaryQueue();
```
wheezil: It would be worth noting that in a more typical use-case, you would persist the reply-to queue to avoid the cost of setup/teardown on each well.

site/jms-client.md
```diff
@@ -367,6 +367,164 @@ may have configuration files or command-line options.
...
+// wait response for 5 seconds
+Message response = queue.poll(5, TimeUnit.SECONDS);
+// close the response consumer
+responseConsumer.close();
+</pre>
+
+Note this sample uses a `MessageListener` and a `BlockingQueue` to wait
+for the response. This implies a network roundtrip to register an AMQP
+consumer and another one to close the consumer.
+`MessageConsumer#receive` could have been used as well, in this case the JMS
```
wheezil: Is receive() advisable in this release?  If it has the 100ms delay issue that would be worth noting here.

site/jms-client.md
```diff
@@ -367,6 +367,164 @@ may have configuration files or command-line options.
...
+            Message response = ...
+            replyProducer.send(replyQueue, response);
+        }
+    } catch (JMSException e) {
+        // deal with exception
+    }
+});
+</pre>
+
+Note the `connectionFactory.setDeclareReplyToDestination(false)`
```
wheezil: :+1:

site/jms-client.md
acogoluegnes: OK, mentioned this solution as welll.

site/jms-client.md
acogoluegnes: Delay issue mentioned.

PR #603, opened 2018-10-17
Correct typo and use only max value for enough traffic

maggyero:


pivotal-issuemaster:
@maggyero Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=603)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=603) to manually synchronize the status of this Pull Request.

site/heartbeats.xml
```diff
@@ -237,9 +237,9 @@ cf.RequestedHeartbeat = 60;
         time against premature closure by proxies and load balancers.
       </p>
       <p>
-        Heartbeat timeouts from 10 to 30 seconds will produce periodic
-        network traffic often enough (roughly every 5 to 15 seconds))
-        to satisfy defaults of most proxy tools and load balancers.
+        Heartbeat timeouts below 30 seconds will produce periodic
```
michaelklishin: This statement is no longer clear. If the timeout is 2 seconds — a value we strongly discourage to use since it's < 5s — then the heartbeat frames will be sent roughly every 1 second. So it should be `of 30 seconds` if the goal is more specificity.

maggyero: @michaelklishin Updated. I found the original sentence ambiguous as it seemed to suggest that timeouts less than 10 second did not produce enough network traffic for proxies and load balancers.

PR #625, opened 2018-11-09
additional clarity for cluster_partition_handling

DaniRey:
Start the chapter with the info, that a cluster might be used for different purposes and therefore has to be tuned accordingly.
Introduce the word split-brain and use it where appropriate for increased clarity (i.e. recovering from a split-brain).
Use datacenter instead of AZ (availability zone), but reference it to have a well known example of the expected network reliability.

I claim that "pause_minority" prevents split-brain during network partitions. In our tests this proofed correct. Please correct me, if this claim is wrong (let's keep the "suspend" an exception, as it is not really the network partition we try to protect against in HA setups).

### DIFF
--- a/site/partitions.xml
+++ b/site/partitions.xml
@@ -24,16 +24,36 @@ limitations under the License.
     <title>Clustering and Network Partitions</title>
   </head>
   <body show-in-this-page="true">
-    <p>
-      RabbitMQ clusters do not tolerate network partitions well. If
-      you are thinking of clustering across a WAN, don't. You should
-      use <a href="/federation.html">federation</a> or
-      the <a href="/shovel.html">shovel</a> instead.
+    <p>RabbitMQ clusters can be used to achieve a higher throughput, 
+      better availability or better data protection.
+      They all require different configurations. In case you plan to spread 
+      your cluster across a WAN, take <a href="/federation.html">federation</a> 
+      and <a href="/shovel.html">shovel</a> into account. They
+      might be the better option for your use-case.
+    </p>    
+    <p>If you run your cluster for better performance, all nodes should remain
+      in the same datacenter, ideally in the same rack. This is the default use-case 
+      and works well with the default configuration. If nothing special is mentioned,
+      the default configuration is assumed in this chapter.
+    </p>
+    <p>If you want to achieve better data protection, you have to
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you. I don't know if "split brain" is any clearer than "network partition" but some parts of this PR are definitely an improvement.

michaelklishin:
Parts of this PR belong to a different guide, Clustering, and to some extent are already covered there. I will edit them out.

DaniRey:
I think split-brain is an improvement. Network partition is what happens in the world, while split-brain is what might happen to the cluster during a network partition.

michaelklishin:
I kept some ideas from this PR and included a round of general edits to the guide. It hasn't seen a serious editing session in a while, so this was a good opportunity. I plan to continue in the upcoming days.

Some additional thoughts: related improvements to the [Clustering](http://www.rabbitmq.com/clustering.html) and [Reliability](http://www.rabbitmq.com/reliability.html) guides are welcome. I'm a stickler for structure in the guides but happy to do extra edits, move things around and reword things as much as necessary.

The Partitions guide could use some more love and this was a great reminder of that. Quorum queues coming in 3.8 will require a non-trivial expansion to it as well.

Thank you, @DaniRey.

michaelklishin:
Forgot the key part: I did leave the split-brain concept part but partition handling is not just split brain handling, so I reverted the title of one section. Nonetheless I hope the concept is covered as @DaniRey intended it to be :)

DaniRey:
I really like the introduction now. My version had too much detail and I struggled to get rid of that.

@michaelklishin I'm happy to provide input and feedback for an update of that part of the documentation. Hit me up via Email or Gitter, during weekends I might respond slowly though.

As RabbitMQ provides a vast amounts of configuration possibilities, I think we could assist the reader a lot, if their would be guides on how to achieve a specific goal. Like increased data safety through replication, increased availability for client operations, higher overall throughput.

I think some graphics might help as well. I some for our documentation. I'm happy to provide them as a blueprint to create them in the same style as the rest of the graphics in the documentation. What would be the best way to get them to you?

@michaelklishin You nailed the split-brain part.

DaniRey:
Hi @michaelklishin 
Tried to reach you via gitter. I do have some ideas for improvement of the documentation.

PR #638, opened 2018-11-25
Initial guide on connections

michaelklishin:
Submitting to shorten the feedback loop on a new guide that covers connections and related topics.


site/connections.md
```diff
@@ -0,0 +1,200 @@
...
+
+
+### High Connection Churn
+
+A system is said to have high connection churn when its rate of newly opened connections is consistently high and
+its rate of closed connection is consistently high. This usually means that an application
+uses short-lived connections. While with some workloads this is a natural state of the system,
+long-lived connections should be used instead when possible.
+
+[Management UI](/management.html) provides a chart of connection churn rate as of [RabbitMQ 3.7.9](/changelog.html).
```
acogoluegnes: This paragraph and the image are duplicated from the previous section.

michaelklishin: New connection rate and connection churn are based on the same metric(s). I did some editing to make it clearer why this chart is relevant for detecting connection leaks.

site/connections.md
```diff
@@ -0,0 +1,261 @@
...
+AMQP 0-9-1 provides a way for connections to multiplex over a single TCP connection. That means an application
+can open multiple "lightweight connections" called channels on a single connection. AMQP 0-9-1 clients open one or more
+channels after [connecting](#lifecycle) and perform protocol operations (manage topology, publish, consume)
+on the channels.
+
+#### AMQP 1.0
+
+AMQP 1.0 provides a way for connections to multiplex over a single TCP connection. That means an application
+can open multiple "lightweight connections" called sessions on a single connection.
+Applications that set up one or more links to publish and consume messages.
```
acogoluegnes: I'm not sure this sentence is correct, I don't understand it.

site/connections.md
```diff
@@ -0,0 +1,261 @@
...
+
+AMQP 0-9-1 supports different authentication mechanisms. While it's most common for applications
+to provide a pair of credentials, x509 certificates and PKI [can be used](https://github.com/rabbitmq/rabbitmq-auth-mechanism-ssl)
+instead.
+
+#### AMQP 1.0
+
+AMQP 1.0 has a model that includes connections, sessions and links.
+
+After successfully opening a connection and authenticating, applications open one or more sessions. It then
```
acogoluegnes: It should be "applications" or "application" in the 2 sentences, but not both?

site/connections.md
michaelklishin: It should have been "then" instead of "that".

PR #640, opened 2018-11-27
Troubleshooting2

gerhard:
This is the outcome of a short discussion with @murphe5 & @ansergit. They were saying that a list of helpful commands when troubleshooting RabbitMQ deployments would go a long way.

There already is a [troubleshooting page](http://www.rabbitmq.com/troubleshooting.html), so maybe it would make sense to work this content in that context. For now, I'm just capturing what was discussed, feel free to take it in any direction that makes sense @murphe5 & @ansergit.

cc @MarcialRosales

### DIFF
--- a/site/troubleshooting2.md
+++ b/site/troubleshooting2.md
@@ -52,3 +52,13 @@ What we actually care about is the following information:
 `rabbitmqctl report`
 
 _We need a better rabbitmqctl report command, the current output is too verbose_
+
+In case `rabbitmqctl list_queues` times out, running these commands will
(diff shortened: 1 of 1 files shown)


michaelklishin:
Let's axe the old existing Troubleshooting guide and start with a new one.

michaelklishin:
I think this should mention that what really narrows problems down is monitoring and data in the opening section.

michaelklishin:
We will reintroduce this guide later with a somewhat different structure based on the recent Monitoring guide improvements and rabbitmq/rabbitmq-cli#292.

PR #642, opened 2018-12-02
Documentation example for rabbitmq/rabbitmq-server#1784

michaelklishin:
See rabbitmq/rabbitmq-server#1784 and rabbitmq/rabbitmq-server#1785.


michaelklishin:
Oops, meant to submit this against `stable`.

PR #666, opened 2019-01-16
Make 3.6 EOL on Erlang Version Requirements page

gerhard:
Submitting as PR in case it needs a discuss. Please enjoy the merge
@michaelklishin if you agree with this change.


michaelklishin:
I can see how this may be valuable for historical purposes. Let's move all EOL series to a separate table. I'll do it.

PR #683, opened 2019-02-11
typed routing key instead of binding key

mamreezaa:
When i went through amqp-concepts it says, 

> when you declare a queue with the name of "search-indexing-online", the AMQP 0-9-1 broker will bind it to the default exchange using **"search-indexing-online" as the routing key.**

If i understood correctly it should be 
> **"search-indexing-online" as the binding key.**

See here : https://www.rabbitmq.com/tutorials/amqp-concepts.html#exchange-default

### DIFF
--- a/site/tutorials/amqp-concepts.xml
+++ b/site/tutorials/amqp-concepts.xml
@@ -167,7 +167,7 @@ limitations under the License.
                  For example, when you declare a queue with the name of
                  "search-indexing-online", the AMQP 0-9-1 broker will bind it
                  to the default exchange using "search-indexing-online" as
-                 the routing key. Therefore, a message published to the default
+                 the binding key. Therefore, a message published to the default
                  exchange with the routing key "search-indexing-online"
                  will be routed to the queue "search-indexing-online". In
                  other words, the default exchange makes it seem like it


pivotal-issuemaster:
@mamreezaa Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=683)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=683) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@mamreezaa Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=683)!

michaelklishin:
Thank you but "routing key" is what the spec and most client libraries use. Both are used interchangeably so I'm afraid we cannot accept this specific change. I've pushed a slightly different one.

mamreezaa:
`public function queue_bind(
        $queue,
        $exchange,
        $routing_key = '',
        $nowait = false,
        $arguments = array(),
        $ticket = null
    )`

i can see it here. 
As long as we understand both terms with the context, it can be interchangeable.

PR #690, opened 2019-02-19
fix documentation

bojanv55:
"but doesn't yet ack it," -> means it is in confirms mode


michaelklishin:
That part looks like an example of overediting: some parts make me think of consumer acknowledgment and not publisher confirms. But your version makes more sense before we can do it a good round of editing 👍

PR #700, opened 2019-03-07
Fix typo in production-checklist.xml

aaabramov:
Or `addressee` was intently used?


michaelklishin:
`addressee` is a word but you are right, the intent was to say `source IP address`. Thank you.

PR #705, opened 2019-03-13
Change http to obs for repositories URL

Gsantomaggio:
Sorry for all these PR(s), but I didn't know about `obs://`. 

`obs://` extracts the http url for the matching distro.
So it is not necessary to specify the distro, this make the life
easier also for automated tools as chef, ansible etc


michaelklishin:
Thank you!

PR #754, opened 2019-05-07
correct missing changes with javascript tutorial codes.

slim-hmidi:
According to #743 there are some code declarations which are not up to date.
I added to missing update in this commit.


michaelklishin:
Thank you!

PR #803, opened 2019-07-17
Add the documentation to run RabbitMQ in ipv6 env

Gsantomaggio:
Complete the documentation for the issue:
https://github.com/rabbitmq/rabbitmq-server/pull/1982


michaelklishin:
This PR needs a lot of editing to make sense to a beginner. It also seems to claim that RabbitMQ tries to "bind to an epmd address/port" which is definitely not the case. It connects to the local epmd. I will edit and merge manually.

PR #808, opened 2019-07-21
remove deprecated stuff in php tutorial

angyvolin:
AMQPChannel::$callbacks was marked as @Internal in php-amqplib/php-amqplib@f30adba and AMQPChannel::is_consuming() public API was introduced to replace it.


michaelklishin:
Thank you!

PR #817, opened 2019-08-05
Ensure to always use pip as a module of a given version of python

4383:
Prefer to use pip as a module, it's avoid error and mistakes during
package installation. With this approach we are always sure to
to install dependencies in the right python environment.


4383:
@lukebakken done, please take a look

PR #827, opened 2019-08-26
Add info on Crash Dumps to Troubleshooting guide

gerhard:
I always find it quicker to navigate Erlang crash dumps with the Viewer app. I think that others will find this helpful as well. What do you think @michaelklishin ?

image


michaelklishin:
Agreed, this is an overdue addition.

PR #855, opened 2019-09-24
Cli docs update

dcorbacho:
Update man pages with the new commands: upgrade, rebalance, consume event stream and command line arguments.


michaelklishin:
@dcorbacho I believe this is hand-crafted instead of produced from the man pages in the RabbitMQ server repository. Let's close this in favor of a separate PR that uses generated pages.

PR #866, opened 2019-10-09
Wrong repo in "Add a Source List File"

XaaT:
Repository is deb https://dl.bintray.com/rabbitmq-erlang/debian $distribution $component
and not deb https://dl.bintray.com/rabbitmq/debian $distribution $component

### DIFF
--- a/site/install-debian.md
+++ b/site/install-debian.md
@@ -192,7 +192,7 @@ pattern:
 <pre class="lang-bash">
 # This repository provides RabbitMQ packages
 # See below for supported distribution and component values
-deb https://dl.bintray.com/rabbitmq/debian $distribution $component
+deb https://dl.bintray.com/rabbitmq-erlang/debian $distribution $component
 </pre>
 
 The next couple of sections discuss what distribution and component values


pivotal-issuemaster:
@XaaT Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=866)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=866) to manually synchronize the status of this Pull Request.

michaelklishin:
The link is not wrong. There is a comment above the line that explains what the repository provides. http://bintray.com/rabbitmq provides RabbitMQ packages while http://bintray.com/rabbitmq-erlang is a separate org that provides Erlang packages only.

The reason for this has been explained on the mailing list: Bintray **completely blocks** free accounts once download traffic goes above a certain threshold and provides no paid plans that would make sense for projects such as RabbitMQ (we don't need most enterprise features  and are not interesting in paying tens of thousands of USD a year to avoid being blocked). So we moved large Erlang packages to a separate org. If that org gets blocked, RabbitMQ downloads are not affected.

callune:
With your explanations it is now clear. But the doc is not as explicit as you are and may be reviewed.
Installation on ubuntu is easy, but when it comes to debian it's not the same...

michaelklishin:
The guide was edited and reviewed many times so I'd like to understand what specifically you found confusing. There is a "copy and paste" shell snippet now which is also annotated. Erlang installation is covered in a separate s ection. I'm not sure how Ubuntu would be different from Debian since Ubuntu's repositories always include an out-of-date RabbitMQ version. If anything, Debian testing repos are typically ahead.

If you have specific recommendations, please post them to the mailing list. We are interested in improving our doc guides but those decisions have to be informed ones.

PR #873, opened 2019-10-19
Add an explanation for default account.

HomoEfficio:


michaelklishin:
Thank you but it feels a bit redundant, two more guides already mentions this (Access Control, Production Checklist). I will add a link to both instead.

PR #894, opened 2019-11-14
Document OAuth 2/UAA support in management plugin

acogoluegnes:
[#169725428]

### DIFF
--- a/site/management.md
+++ b/site/management.md
@@ -291,21 +291,20 @@ management.uaa_client_id = rabbit_user_client
 management.uaa_location = https://my-uaa-server-host:8443/uaa
 </pre>
 
-When using `enable_uaa = true`, it is still possible to authenticate
+When using `management.enable_uaa = true`, it is still possible to authenticate
 with [HTTP basic authentication](https://en.wikipedia.org/wiki/Basic_access_authentication)
-against the HTTP API. This means the following:
+against the HTTP API. This means both of the following examples will work:
 
 <pre class="lang-bash">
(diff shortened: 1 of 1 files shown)


dcorbacho:
@acogoluegnes Should we show/explain the redirect and logout? Also how tokens are sent on the HTTP API?

acogoluegnes:
> @acogoluegnes Should we show/explain the redirect flow and logout?

Not sure it's necessary, the page is already quite long.

> Also how tokens are sent on the HTTP API?

Can you be more specific? I don't remember there's something special regarding the HTTP API.

dcorbacho:
Almost all requests use in the authorization header `Bearer <token>' as expected, but the `definitions` import/export endpoint needs to use `token=<token>` in the query string.

acogoluegnes:
> Almost all requests use in the authorization header `Bearer <token>` as expected, but the `definitions` import/export endpoint needs to use `token=<token>` in the query string.

This is when HTTP basic authentication is disabled, right? This is worth mentioning indeed, I'll add it.

site/management.md
```diff
@@ -275,6 +275,82 @@ rabbitmqctl set_user_tags full_access administrator
 </pre>
 
 
+## <a id="http-api" class="anchor" href="#http-api">Authenticating with OAuth 2</a>
```
michaelklishin: The anchor here is a copy-paste artifacts and it break on page navigation.

acogoluegnes: Good catch, fixed.

PR #951, opened 2020-03-06
Align syntax for rabbitmq-diagnostics

oneoneonepig:


michaelklishin:
Thank you for your time but I don't see how this change is meaningful.

oneoneonepig:
The rest of the command examples are in the syntax of `rabbitmq-diagnostics -q <command>`, just like what written in the man page and the help page. I know that the cli will parse the options properly, just like typing `yum install -y rabbitmq` and `yum install rabbitmq -y`, but during first time I read through the documents, I feel like the option '-q' is for 'ping', not for 'rabbitmq-diagnostics'.

At least this is meaningful for me, in terms of user experience. If that makes no sense for you, you can leave it as is.

Cheers

michaelklishin:
Thank you!

PR #962, opened 2020-04-12
Fixed markup in mqtt.md

lyrixx:


michaelklishin:
I'm not sure why the block wouldn't format correctly (GHFM allows for whitespace between the ``` and the language identifier to my knowledge) but OK.

michaelklishin:
Thank you!

PR #977, opened 2020-05-30
The output of the sender has error

steven-wan:
when  if (dots.incrementAndGet() == 3)  dots.set(1), the value is 1,not 3,
so for (int i = 0; i < dots.get(); i++) { builder.append('.'); } can not println hello...   
it reality println
 [x] Sent 'Hello.1'
 [x] Sent 'Hello..2'
 [x] Sent 'Hello.3'
 [x] Sent 'Hello..4'
 ......


steven-wan:
incrementAndGet() return the updated value , so  if (dots.incrementAndGet() == 3)  true , dots.set(1).

michaelklishin:
Thank you!

PR #994, opened 2020-06-25
Kubernetes clustering: refer to statefulset instead of deployment

ferozjilla:
There is a general recommendation in the docs to use stateful sets for deploying RabbitMQ on Kubernetes. 

The linked example in the `rabbitmq-peer-discovery-k8s` repository also defines a statefulset (although the file right now is called deployment.yaml).


michaelklishin:
Thank you!

PR #1035, opened 2020-08-14
Simplified installation process for the Kubernetes Operator

mkuratczyk:
* delete helm chart method
* remove unnecessary steps (thanks to https://github.com/rabbitmq/cluster-operator/pull/250)
* group all commands together so that you can just copy-pasting all
commands at once should work for most people
* move image relocation and private registries to the bottom of the page
as most people don't need them

This PR assumes https://github.com/rabbitmq/cluster-operator/pull/250 is merged (it should be today). Until it is, you can test the installation steps by adding `git co simplified-installation` after cloning `cluster-operator` repo


mkuratczyk:
https://github.com/rabbitmq/cluster-operator/pull/250 has now been merged so no need to switch to the branch

michaelklishin:
Thank you, this is definitely an improvement!

PR #1047, opened 2020-08-31
Add Foo-Foo-MQ to devtools for node

auroq:
@zlintz and I have recently taken over the decently popular node library [Rabbot](https://github.com/arobson/rabbot) (with the support of its owners).

We have re-branded the library as [foo-foo-mq](https://github.com/foo-foo-mq/foo-foo-mq) and maintain it and some related dependencies under the organization [Foo-Foo-MQ](https://github.com/Foo-Foo-MQ/).

We recently released version 5.0.0 with support for node 14 and amqplib 6 and are averaging between one and two thousand downloads a week at this point.

We would love it if we could be added to the list on this page. Thank you!

### DIFF
--- a/site/devtools.md
+++ b/site/devtools.md
@@ -105,6 +105,8 @@ Miscellaneous projects:
  * [amqp-stats](https://github.com/timisbusy/node-amqp-stats): a node.js interface for RabbitMQ management statistics
  * [Rascal](https://github.com/guidesmiths/rascal): a config driven wrapper for [amqp.node](https://github.com/squaremo/amqp.node) supporting multi-host connections,
    automatic error recovery, redelivery flood protection, transparent encryption and channel pooling.
+ * [foo-foo-mq](https://github.com/foo-foo-mq/foo-foo-mq): an opinionated abstraction over [amqp.node](https://github.com/squaremo/amqp.node)
+   that uses reasonable assumptions and defaults to simplify the implementation messaging patterns on RabbitMQ.
 
 
 ## <a id="objc-swift-dev" class="anchor" href="#objc-swift-dev">Objective-C and Swift</a>


pivotal-issuemaster:
@auroq Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=1047)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=1047) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@auroq Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-website&pullRequestId=1047)!

michaelklishin:
Thank you for taking over an open source project and taking the time to contribute! We usually add projects to that page when they pick up a certain amount of traction. I only remember Rabbot mentioned once on our mailing list or Stack Overflow. I suggest that you announce the renaming on the mailing list and in RabbitMQ community Slack first.

It's entirely up to you how you decide to name the library but I feel that "foo-foo-mq" is a counterproductive name. For starters, it sounds like a joke and not a "serious" project. Unfortunately, such things do hurt adoption. Honestly, I don't know if I would recommend "foo-foo-mq" to a user or prospective customer as an "easier option to amqplib" with a straight face. Could be just me, of course.

Second, it makes it sound like the project is a messaging broker and not an advanced client or messaging framework )that may be protocol-independent or target one of the protocols that RabbitMQ supports and its features).

Some names that I like in a similar space? MassTransit. Hutch. EasyNetQ is OK (at least it communicates its original goal right in the name). Sneakers tries to communicate convenience and I cannot think of any negative connotations.

gerhard:
I am curious about the name too. I've read it `Foo-Bar-MQ` initially, and `Foo-Foo-MQ` on the second pass. I cannot help think of [Boaty McBoatface](https://en.wikipedia.org/wiki/Boaty_McBoatface) and smile. It's an interesting name with an even interesting history, and I'm wondering if the same applies here.

As a separate thought, I am wondering if the foo-foo-mq abstractions shouldn't be contributed to [amqplib](https://github.com/squaremo/amqp.node) directly. I am not seeing anything too controversial. I'm wondering if by working through integrating them in an already popular Node.JS AMQP 0-9-1 library it wouldn't be more beneficial to the node.js RabbitMQ users. I know for a fact that amqlib would appreciate a different perspective. Who knows, some things may improve within the library as a result of this potential collaboration.

I am all for forking and building on top, but [this XKCD](https://imgs.xkcd.com/comics/dependency_2x.png) makes me think that in some cases consolidation is more beneficial than divergence. Yes, it's harder, I won't pretend otherwise, but it takes a village.

michaelklishin:
As the README states, this project is opinionated. We would not accept MassTransit or EasyNetQ abstractions and choices into the RabbitMQ .NET client. NServiceBus goes even further than those two. So it should be a separate project, just named differently.

PR #1052, opened 2020-09-10
Remove SpringOne banner

jacksoncvm:


michaelklishin:
@jacksoncvm thank you. Should we put a VMworld one in its place?

PR #1063, opened 2020-09-28
Do not mention RabbitMQ 2.x

mkuratczyk:
* I think (certainly hope!) we don't need to mention RabbitMQ 2.x anymore
* in a similar vain, I mentioned that Erlang 19 is very old

### DIFF
--- a/site/upgrade.md
+++ b/site/upgrade.md
@@ -409,10 +409,6 @@ which disc node will be the upgrader, stop that node last, and start it first.
 Otherwise changes to the cluster configuration that were made between the
 upgrader node stopping and the last node stopping will be lost.
 
-Automatic upgrades are only possible from RabbitMQ versions 2.1.1 and later.
-If you have an earlier cluster, you will need to rebuild it to upgrade.
-
-
 ## <a id="caveats" class="anchor" href="#caveats">Caveats</a>
 
 There are some minor things to consider during upgrade process when stopping and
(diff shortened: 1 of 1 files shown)


site/upgrade.md
```diff
@@ -426,7 +422,7 @@ during shutdown, which blocks subsequent upgrade steps:
 * [OTP-14441](https://bugs.erlang.org/browse/ERL-430): fixed in Erlang/OTP `19.3.6` and `20.0`
 * [OTP-14509](https://bugs.erlang.org/browse/ERL-448): fixed in Erlang/OTP `19.3.6.2` and `20.0.2`
 
-A node that suffered from the above bugs will fail to shut down and stop responding to inbound
+Please note that both issues affect very old version of Erlang (check [Erlang Version Requirements](https://www.rabbitmq.com/upgrade.html#rabbitmq-erlang-version-requirement) for more details). A node that suffered from the above bugs will fail to shut down and stop responding to inbound
```
michaelklishin, line 425: Should we link directly to `/which-erlang.html` here?

michaelklishin (changes requested):

michaelklishin:
Since this is a guide on upgrading, it's reasonable to expect Erlang 19 users in the audience. I don't think I have seen anyone running RabbitMQ older than 2.7 in many years, and even 2.7 was an outlier.

PR #1083, opened 2020-11-09
Rename resources and document some new k8s rabbitmqcluster configurations

ChunyiLyu:


yaronp68 (changes requested):
I’m ok with the naming changes. Can we remove the word pivotal from the examples? Also can we call rabbitmq management image, rabbitmq with management plugin image?

yaronp68 (approved):

PR #1085, opened 2020-11-10
Fix service account name to have suffix '-server'

ansd:
Since https://github.com/rabbitmq/cluster-operator/pull/447 is merged, the service account name is `<instance-name>-server`.


Zerpet (approved):

ChunyiLyu (approved):
Great!

PR #1089, opened 2020-11-11
WIP Remove caCertName from spec.tls

ablease:
This is now hardcoded to 'ca.crt'.

Context: https://github.com/rabbitmq/cluster-operator/issues/267
We decided that this property name was confusing, and have opted to
remove it to keep TLS more conventional.

Please do not merge this to live until we have completed this issue in the operator (see link above)

I will update this PR when are ready.


ChunyiLyu:
@ablease I think changes in the PR is already in: https://github.com/rabbitmq/rabbitmq-website/blob/live/site/kubernetes/operator/using-operator.md#tls-conf

Could you confirm and close the PR if that's true?

PR #1091, opened 2020-11-14
Fix anchor in dlx.md

GitHubPang:
Correct link anchor (was duplicated with the one on line 158).


michaelklishin (approved):

michaelklishin:
Thank you!

PR #1109, opened 2020-12-08
Add missing rabbitmq-upgrade commands

mkuratczyk:
await_online_quorum_plus_one, await_online_synchronized_mirror, drain and revive were missing


michaelklishin:
Thank you. Good catch. These files are generated from [man page gross/troff sources](https://github.com/rabbitmq/rabbitmq-server/tree/master/deps/rabbit/docs) and since most people would rather eat a cactus than figure out troff format, I'll update them on your behalf and re-sync with the site :)

PR #1153, opened 2021-02-25
Explain that prefetch of 0 is treated as infinite

nathanl:
See https://github.com/rabbitmq/rabbitmq-server/blob/17b082abeb19340eea98a9700b32a964db75dadc/deps/rabbit/docs/rabbitmqctl.8#L1051-L1054


michaelklishin (approved):

michaelklishin:
Thank you!

PR #1160, opened 2021-04-05
Add SSL password option to Management Plugin docs page

2YY:
Hi.

I'm tried to connect Management UI with HTTPS, finally I realized my `rabbitmq.conf` lacks `management.ssl.password = bunnies` and  worked fine by adding it to my `rabbitmq.conf`. ([Thanks Michał!](https://groups.google.com/g/rabbitmq-users/c/7hbvKq_Xdc4))

Then, I added it also to document for RabbitMQ beginners like me. And created this PR.

Thanks.

* note: Similarly, I should also add `management.ssl.verify = verify_peer` and `management.ssl.fail_if_no_peer_cert = true` to the Management Plugin docs page?

### DIFF
--- a/site/management.md
+++ b/site/management.md
@@ -421,6 +421,7 @@ management.ssl.port       = 15671
 management.ssl.cacertfile = /path/to/ca_certificate.pem
 management.ssl.certfile   = /path/to/server_certificate.pem
 management.ssl.keyfile    = /path/to/server_key.pem
+management.ssl.password   = bunnies
 </pre>
 
 More [TLS options](/ssl.html) can be configured for the HTTPS listener.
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you. I will merge this manually with some edits. We cannot just add every single option to the example as most users won't necessarily need them. We can mention the peer verification options but it's actually relatively rare for HTTP API servers to authenticate clients (e.g. GitHub does not perform peer verification on your browser but your browser does for the chain that github.com serves). But it can be mentioned with a link to [this main TLS guide section](https://www.rabbitmq.com/ssl.html#peer-verification).

michaelklishin:
I will add those options on your behalf.

michaelklishin:
Apparently the example that follows the one modified in this PR already mentions `management.ssl.password`.

2YY:
Thank you for your quick and polite response!

PR #1161, opened 2021-04-05
Add local-host provisioner in kubernetes quickstart

rikatz:
Currently quickstart does not cover a local provisioner.

This might lead to bad user experiences, as nowhere in the docs there's a mention of the need of a Volume provisioner :)

So the idea here is, if the user is stuck because of PVC, add a local path provisioner just to make sure he can follow with the quickstart guide :D


rikatz:
Other idea for this is maybe move to troubleshooting and explain that the operator relies on a Storage Class provisioner :D

michaelklishin:
Thank you!

PR #1176, opened 2021-05-07
Fix typo & missing link in debian install doc

coro:
I wasn't sure where the empty link was intended to point to, so I picked a likely destination. Feel free to propose an alternative, or I can revert the link to point back to the top of this doc.


acogoluegnes (approved):

michaelklishin:
You've picked correctly, thank you :)

PR #1178, opened 2021-05-07
Update news.xml

etc-crontab:
Fix the release version to target for version 3.8.15.


michaelklishin:
Good catch, thank you.

PR #1183, opened 2021-05-14
added .tut1 to package in rabbitmq springboot tutorial

BarbosaRicardo:
While following the first tutorial for hello-world in spring boot, I noticed that a package name was missing 'tut1' at the end.


michaelklishin:
Fair enough, thank you.

PR #1193, opened 2021-05-27
Fix Vhost kind in yaml example for topology operator documentation

senguehard:
In the documentation for using the topology operator, the Vhost example has a `kind: Queue` instead of a `kind: Vhost` in the yaml example. This PR fixes that, according to the example in https://github.com/rabbitmq/messaging-topology-operator/blob/main/docs/examples/vhosts/vhost.yaml


michaelklishin (approved):

michaelklishin:
Thank you!

PR #1197, opened 2021-06-04
Fixed broken kubernetes operator link

Xilis:
Reading up on the k8s operator I noticed that this link is broken.

Instead of linking to https://kubernetes.io/docs/concepts/extend-kubernetes/operator/ it links to https://www.rabbitmq.com/kubernetes/operator/(https://kubernetes.io/docs/concepts/extend-kubernetes/operator/) which is a broken link.


michaelklishin (approved):

michaelklishin:
Thank you!

PR #1198, opened 2021-06-07
Remove randomized startup delays

ansd:
This is the doc change for https://github.com/rabbitmq/rabbitmq-server/pull/3075 (i.e relevant for `>= 3.8.18`)

In K8s, the cluster-operator nowadays uses Parallel (instead of OrderedReady)
pod management policy. Therefore, we delete some sentences on
recommending OrderedReady.

There is no need to document the new config value
`cluster_formation.internal_lock_retries` since it's too much
implementation detail for this doc and we don't expect users to change
this value.


michaelklishin:
I reworded this a bit, looks good to go after `3.8.17` ships.

michaelklishin:
Backported to `v3.8.x`.

PR #1200, opened 2021-06-07
Document design principles of Cluster Operator

Zerpet:
Based on issue https://github.com/rabbitmq/cluster-operator/issues/712, I noticed we don't have any written reference of our design principles, and this can cause confusion, as there isn't an standarised behaviour in the Operator ecosystem.

These ideas/principles have been present in our decisions since the beginning, however there were not written anywhere. This PR flushes them in the Operator Overview page.


Zerpet:
@ChunyiLyu and/or @mkuratczyk, can you review this please? 🙏

mkuratczyk:
The order of sentences makes it sound like not upgrading instances is a side-effect of re-populating the `image` field when it is removed (rather than of not modyfying existing specs). Other than that, looks good to me

mkuratczyk (approved):

michaelklishin:
Reworded a bit and added a link to the configuration guide.

PR #1203, opened 2021-06-09
Update using-topology-operator.md TOC to link properly

cassid4:
Fix table of contents to reference correct header IDs


michaelklishin:
Thank you!

PR #1206, opened 2021-06-10
add paths for configuration files to documentation

nick-brady:
Hello,

I was trying to open my rabbitmq server on all interfaces (0.0.0.0) in order to test out something with another device on my network. I was confused why it wasn't working. When I searched around I found [a GitHub gist](https://gist.github.com/robcowie/1317580) which finally told me the path of the configuration file where I changed `NODE_IP_ADDRESS` from `127.0.0.1` to `0.0.0.0` and it worked. When I was reading the documentation, I thought this page would be a helpful place to put this information.

I won't be offended if language is changed at all, but am happy to do it myself as well.

Nick


michaelklishin:
Thank you. There is already a dedicated section [in the Configuration guide](https://www.rabbitmq.com/configure.html#config-location). You are welcome to submit a PR that links to it.

PR #1209, opened 2021-06-15
Updated Rabbitmqctl man page with correct example for 'set_log_level'

jaisea:
The example command for "set_log_level" was wrong. Updated Rabbitmqctl man page with correct example for 'set_log_level'


michaelklishin:
Thank you but this file is not the source. The source is in the Troff format so I'll update it on your behalf :)

michaelklishin:
I will regenerated and deploy this change in a bit, apparently more changes were committed directly to the HTML version
and have to be backported.

michaelklishin:
Done and going live in a few minutes.

PR #1213, opened 2021-06-25
Update README master references to main

g1stavo:
README was still referencing `master` branch, which is now `main`


michaelklishin (approved):

michaelklishin:
Good catch, thank you.

PR #1216, opened 2021-06-26
Fix incorrect wording

GitHubPang:
Fix incorrect wording concerning the original name of "queue leader".


michaelklishin:
This is not "incorrect wording". We avoid using "master" where possible.

michaelklishin:
Ah, I see there was one replacement too many.

PR #1218, opened 2021-06-27
Restore documentation on `reject-publish-dlx` overflow setting

### LINKED ISSUE #1217: Missing documentation on `reject-publish-dlx` overflow setting
Looks like the patch of #774 was lost during conflict resolution in 3cb9d3156f0de7109ba33195379a3fd81547d888.

Invizory:
This is the same change as in #774, expect that Markdown is used.

See rabbitmq/rabbitmq-server#1443, rabbitmq/rabbitmq-server#2015.

Closes #1217.


michaelklishin:
Thank you!

PR #1226, opened 2021-07-10
Update tutorial-six-javascript.md

shreyshreyansh:
Grammatical error


deadtrickster:
Quoting this: https://english.stackexchange.com/a/56911

>Contrast this with the initialism "RPM", which begins with the consonant "R" but is pronounced starting with /a/; thus, you use an, as in an RPM counter.

So it's indeed "an RPC"

michaelklishin:
It's a  matter of style and depends on how the abbreviation is typically pronounced. 

This [APA Style blog post](https://blog.apastyle.org/apastyle/2012/04/using-a-or-an-with-acronyms-and-abbreviations.html) can be of use. In the docs we assume that "RPC" is usually pronounced as `[ˌɑːpˌiːsˈiː]` ("arr-pee-see").

PR #1233, opened 2021-07-23
Add instruction to place ca cert in webhook configurations

ChunyiLyu:
This PR adds instruction to place ca cert in webhook configurations. This is a necessary step for the installing topology operator with generated certs. This documentation problem is reported by @evankanderson in issue: https://github.com/rabbitmq/messaging-topology-operator/issues/185

### DIFF
--- a/site/kubernetes/operator/install-topology-operator.md
+++ b/site/kubernetes/operator/install-topology-operator.md
@@ -58,6 +58,32 @@ data:
   ca.crt: # ca cert that can be used to validate the webhook's server certificate
   tls.crt: # generated certificate
   tls.key: # generated key</pre>
+3. Add webhook ca certificate in downloaded release manifest `messaging-topology-operator.yaml`. There are 10 admission webhooks, one for each CRD type.
+   Look for keyword `clientConfig` in the manifest, and paste the webhook ca cert under `clientConfig.caBundle`.
+   Because there are 10 different webhooks, you need to perform this action in 10 different places.
+   The example below shows how to add ca cert to the `queues.rabbitmq.com` validating webhook:
+<pre class="lang-yaml">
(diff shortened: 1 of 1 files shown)


site/kubernetes/operator/install-topology-operator.md
```diff
@@ -58,6 +58,32 @@ data:
   ca.crt: # ca cert that can be used to validate the webhook's server certificate
   tls.crt: # generated certificate
   tls.key: # generated key</pre>
+3. Add webhook ca certificate in downloaded release manifest `messaging-topology-operator.yaml`. There are 10 admission webhooks, one for each CRD type.
```
michaelklishin, line 61: If there are 10 of them, should this say "CA certificates"?

MirahImage (approved):
It's probably non-trivial to automate, but we should keep in mind that we'll need to update the number of places the certificate is referenced any time we add another kind to the operator.

site/kubernetes/operator/install-topology-operator.md
ChunyiLyu: What I was trying to express is add the same generated certificate to 10 different places. I can certainly word it differently to avoid confusion here.

michaelklishin: Then should it say "Add the web hook CA certificate found in the downloaded release manifest, `messaging-topology-operator.yaml`"?

michaelklishin:
Merged manually.

site/kubernetes/operator/install-topology-operator.md
evankanderson: I think the `clientConfig` hint here is good -- v0.8.0 also included CRD conversion webhooks in addition to the validating admission webhooks. (It seems like the conversion webhooks were removed in later versions).

PR #1235, opened 2021-07-30
Make explicitly clear the precedence of RABBITMQ_LOGS

Zerpet:
We recently received a question on how to configure logging to a file in the Cluster Operator:

https://github.com/rabbitmq/cluster-operator/issues/771

The conclusion was that the Docker image sets RABBITMQ_LOGS environment variable, therefore any configurations like `log.file` were simply ignored. To me, it was not clear that the environment variable is not additive. A user could think that setting `RABBITMQ_LOGS` to `-` and `log.file` to a path would log to both, standard output and a file. This is not true and we can be explicitly clear about it in the docs.


michaelklishin (approved):

michaelklishin:
Thank you!

PR #1240, opened 2021-08-10
Update Debian GPG instructions to bring inline with best practice

udf2457:
As per https://wiki.debian.org/DebianRepository/UseThirdParty

"The key MUST be downloaded over a secure mechanism like HTTPS to a location only writable by root, which SHOULD be /usr/share/keyrings. The key MUST NOT be placed in /etc/apt/trusted.gpg.d or loaded by apt-key add. A sources.list entry SHOULD have the signed-by option set. The signed-by entry MUST point to a file, and not a fingerprint."

Should you require further background beyond the above Debian wiki link, see also:
[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=861695
[2] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=877012

### DIFF
--- a/site/install-debian.md
+++ b/site/install-debian.md
@@ -165,16 +165,15 @@ All steps covered below are **mandatory** unless otherwise specified.
 
 ### Add Repository Signing Key
 
-In order for `apt` to use the repository, [RabbitMQ signing key](signatures.html) must be added to the list of trusted keys.
-This will instruct apt to trust packages signed with that key.
+In order for `apt` to use the repository, [RabbitMQ signing key](signatures.html) must be available to the system for validation.
 
 <pre class="lang-bash">
 ## Team RabbitMQ's main signing key
-sudo apt-key adv --keyserver "hkps://keys.openpgp.org" --recv-keys "0x0A9AF2115F4687BD29803A206B73A36E6026DFCA"
+curl -1sLf "https://keys.openpgp.org/vks/v1/by-fingerprint/0A9AF2115F4687BD29803A206B73A36E6026DFCA" | gpg --dearmor > /usr/share/keyrings/com.rabbitmq.team.gpg
 ## Launchpad PPA that provides modern Erlang releases
-sudo apt-key adv --keyserver "keyserver.ubuntu.com" --recv-keys "F77F1EDA57EBB1CC"
+curl -1sLf "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0xf77f1eda57ebb1cc" | gpg --dearmor > /usr/share/keyrings/net.launchpad.ppa.rabbitmq.erlang.gpg
 ## PackageCloud RabbitMQ repository
-curl -1sLf 'https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey' | sudo apt-key add -
+curl -1sLf "https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey" | gpg --dearmor > /usr/share/keyrings/io.packagecloud.rabbitmq.gpg
 </pre>
 
 See the [guide on signatures](signatures.html) to learn more.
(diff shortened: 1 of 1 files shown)


michaelklishin:
Intresting. Thank you!

michaelklishin:
This broke site generation because `&` and `>` have to be replaced with their HTML/XML escape sequences but I understand that it is not obvious and will correct.

udf2457:
@michaelklishin Thanks & sorry about that !

michaelklishin:
No worries. I learned a lot after reviewing this. The instructions do look more involved now but no objections from me.

@udf2457 this will go live in a few minutes after I push my escaping fixes. I have noticed that there are `.list` file snippets in the guide that were not updated. Should they? or would it not make much sense to the reader unless they keep in mind how/where the key was installed?

michaelklishin:
There are also [apt instructions in the Signatures guide](https://www.rabbitmq.com/signatures.html). Ideally we should use the same
approach across the board, even if most people will use the already updated snippet. This may need a bit of wording changes to explain
what those lines do, e.g. as inline `.list` file comments. WDYT?

udf2457:
@michaelklishin I think you're right, they should have been updated.  By the same token (unless I'm misreading the page) I get the impression there is also quite a lot of repetitive duplication going on that page ?  Might be time for some extravagant use of the delete key for certain sections ?

Good spot on the signatures guide, should be updated too as you say in order to keep inline.

udf2457:
@michaelklishin Also, if we are going to be doing more doc updates in this area, then it would be neat if you could ensure binary pubkeys are available everywhere.  This would:

1. Ensure things remain inline with Debian best practice (_"The reason why we avoid ASCII-armored files is that they cannot be used directly by SecureApt."_ from the Wiki)
2. Would remove/reduce the need to pipe to `gpg --dearmor` in the docs and so keep things neater.

i.e. https://github.com/rabbitmq/signing-keys/releases and on packagecloud too.

michaelklishin:
We can put out a new "release" in `rabbitmq/signing-keys` but we do not control PackageCloud or Cloudsmith keys. They are provided by those services.

udf2457:
So `https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey` is not uploaded by you ? (Apologies, I'm not familiar with the workings of packagecloud, have only ever been a consumer of the service, never a publisher on there).

michaelklishin:
@udf2457 there is some repetition because we cover multiple package services. Let's leave them as is and focus on hardening the part related to keys. We would be happy to produce a binary version of the key but it's not something I can promise to happen in a week or two in a high PTO season. I can merge a reasonable docs PR very quickly, though ;)

michaelklishin:
@udf2457 correct, PackageCloud generates a key per repository (used to be a key per org account IIRC). We do not control it.

Cloudsmith allows for custom signing keys but OSS plans do not include this feature, so we rely on GPG and RSA keys generated by the service and only can regenerate it at any time.

udf2457:
@michaelklishin Thanks for the clarifications.

What are your plans for a Bullseye Debian build given it was released yesterday ? (https://www.debian.org/News/2021/20210814)

Maybe I should wait for that to happen before looking at taking a bit more time to update the remaining bits of the docs ? WDYT?

michaelklishin:
Our Debian package should work just fine on Bullseye. The package is the same for all distributions. Our Erlang package has to be adapted but last time we tried it failed to build on Bullseye for reasons we could not understand.

michaelklishin:
I don't see any reason to provide Bullseye-specific docs immediately. I doubt they will be meaningfully different.

Bullseye ships with [Erlang 23.2](https://packages.debian.org/search?keywords=erlang-nox&searchon=names&suite=all&section=all) which is sufficient for running RabbitMQ through mid-2022 (when Erlang 25 will be released), so we are not in rush to produce our Debian packages for Bullseye.

udf2457:
Have posted update for signatures page, will work on the Debian page separately.

Fun fact, I guess nobody uses the signatures page as you had quite a major typo in the old instructions `apt-key adv`  (adv vs add) !

PR #1241, opened 2021-08-11
Update rabbitmqctl.8.html

lushndm:
Command 'list_topic_user_permissions' not found. 
Did you mean 'list_user_topic_permissions'?


michaelklishin:
The source of these man pages is in the server repository. I will update it. Thanks.

PR #1243, opened 2021-08-16
Remove SKS + Add Debian Best Practices

udf2457:
1. Remove all references to SKS which is now dead (https://code.firstlook.media/the-death-of-sks-pgp-keyservers-and-how-first-look-media-is-handling-it)
2. Debian best-practice as per https://github.com/rabbitmq/rabbitmq-website/pull/1240


udf2457:
Ping #1240

PR #1244, opened 2021-08-16
Debian best-practice + reduce verbosity

udf2457:
1. Debian best-practice as per https://github.com/rabbitmq/rabbitmq-website/pull/1240
2. Make a start on removing some of the excessive verbosity on the page (e.g. one example for `$distribution` substitution is sufficient)


udf2457:
Just as an FYI on (2).  Your comment on #1240 ("repetition because we cover multiple package services") is noted and respected.  However the page as-is is very very long, and I think at least some of the low-hanging fruit should be addressed, which is what I have (partially) done today (i.e. reducing three examples of how to use the distro table to one).

PR #1246, opened 2021-08-16
Simplify wording re: Supported Distributions + PackageCloud one-liner note

udf2457:
Remove verbose discussion of OpenSSL versions etc. in favour of the easier to understand "current and previous distro version".

Also add note re: PackageCloud one liner.


michaelklishin:
Thanks. Our support timeline is close to that of Debian releases so we can adopt a shorter version.

PR #1252, opened 2021-08-16
Ordering tweaks + remove "older linux distributions"

udf2457:
1. Removed "Without systemd (Older Linux Distributions)"
2. Tweak PackageCloud section ordering
3. Tweak Cloudsmith ordering
4. Remove `apt-key` references from signatures page


udf2457:
Thanks @michaelklishin I think that's enough fun for one day .... I'll (try to) leave you in peace for a couple of days or so !

PR #1254, opened 2021-08-17
insert sudo's in signature page

udf2457:


udf2457:
I should get back to real $work  ;-)

michaelklishin (approved):

PR #1260, opened 2021-08-20
Move Openshift instructions to its own page based upon user feedback

coro:


site/kubernetes/operator/using-operator.md
```diff
@@ -22,7 +22,6 @@ This guide is structured in the following sections:
 * [Monitor RabbitMQ Clusters](#monitoring)
 * [Delete a RabbitMQ Instance](#delete)
 * [Pause Reconciliation for a RabbitMQ Instance](#pause)
-* [Using the Cluster Operator on OpenShift](#openshift)
```
michaelklishin, line 25: If this moves to a separate guide, should we still link to it? I don't see any new links in this PR.

coro: We still link to it where it's relevant, i.e. creating a RabbitmqCluster: https://github.com/rabbitmq/rabbitmq-website/pull/1260/files#diff-b3d8606acd9b562a9760ce5f19a11c93453127bfb493cd79fa4171e97299192cR104

michaelklishin: Yes but only to a specific section from a specific section. Sounds like it should be a part of the general ToC and "Operator guide ToC" which we have at the top.

In other words, it should be as easily discoverable as other Operator guides.

michaelklishin (approved):

PR #1262, opened 2021-08-21
Update networking.md default port

Yousiph1:
The is a mistake on the rabbitmq default port, it is 15672, not port 5672.


michaelklishin:
It [depends on what protocol we are talking](https://www.rabbitmq.com/networking.html#ports).

Yousiph1:
Yeah, I just realized port 15276  if for the management plugin

PR #1263, opened 2021-08-24
Fixing some of the language a little

thesteve0:
Just making the gender neutral stuff (which I applaud) a bit less awkward.


michaelklishin (approved):

michaelklishin:
Thank you!

PR #1275, opened 2021-09-21
Add cluster size recommendations for performance and throughput

dreiucker:
## Purpose of this PR

Add specific cluster size recommendations regards performance and throughput look like

## Reason for such a change

We are having lengthy discussions with stakeholders how a proper RabbitMQ cluster topology looks like. Since we assume that it would help our service as well as other RabbitMQ users if RabbitMQ could specific that area in the official RabbitMQ documentation instead of having it indirectly documented via several blog posts or deeper understanding of the cluster architecture. 

## Why a PR?

This is just a recommendation what we would add. And since it was mentioned in slack that PR > Issue I followed that guideline. 

## Slack discussion
https://rabbitmq.slack.com/archives/C1EDN83PA/p1630339373040200

### DIFF
--- a/site/production-checklist.md
+++ b/site/production-checklist.md
@@ -392,6 +392,11 @@ both during operations and node restarts. Users who find themselves in need of
 clusters with node counts in double digits should
 **consider using independent clusters for separate parts of the system** where possible.
 
+From a performance standpoint, the maximum nodes recommended is 9 nodes in one RabbitMQ cluster.
+If you are attempting to increase performance and throughput,
+anything in the double digit node counts is not the solution.
+Instead, it is optimal to create new RabbitMQ cluster to improve messaging performance
+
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you for taking the time to contribute! My discussion with the team suggests we can accept this particular wording. There are aspects in which having double digit nodes would be problematic until we replace the schema data store. In other ways, it can be a valid approach.

We'd really want to avoid blanket recommendations on this very broad topic in general.

dreiucker:
@michaelklishin I'm getting confused. You wrote that "team suggests we can accept this particular wording" , but the PR was closed. Do you rather mean that you **can't** accept the PR?

michaelklishin:
Yup, it was a typo, sorry. We cannot accept it.

PR #1279, opened 2021-09-23
Correct consumer timeout

ansd:
It seems that no consumer timeout was the default a while ago.

Relates https://github.com/rabbitmq/rabbitmq-server/pull/2990


michaelklishin (approved):

michaelklishin:
Good catch, thank you!

PR #1284, opened 2021-10-04
Using 'tee' command instead of '>'

sergei66666:
There is error 'Permission denied' at least when install rabbitmq on Ubuntu 20.04. If we use 'tee' all works fine.

(thanks to https://stackoverflow.com/a/82553/9252162)


michaelklishin (approved):

michaelklishin:
Redirection worked for me when I tried it but fair enough, we can switch to using `tee` I guess.

sergei66666:
@michaelklishin may be you was like Root user? I assume in that case redirection work.

PR #1288, opened 2021-10-08
Correct rabbit.default_consumer_prefetch example

lewishenson:
The current example does not work and results in an error when trying to open a channel ("no match of right hand side value").


michaelklishin:
Good catch, thank you.

PR #1297, opened 2021-11-03
Add Delivery-Limit Dead-Lettered Reason

EddieChoCho:
If a message is dead-lettered after it has been returned more times than the limit, the `reason` field of the `x-death` header will be "delivery-limit".


michaelklishin (approved):

michaelklishin:
Thank you!

michaelklishin:
@pivotal-cla obvious fix

PR #1300, opened 2021-11-12
Add 3.8.25

johanrhodin:


Zerpet:
This change was introduced in the [live site](https://rabbitmq.com/which-erlang.html), via ed4f6e8669c981ce1dbfc0a8331116bb28e4424c. I believe this PR will be a no-op.

PR #1308, opened 2021-11-29
Fix invalid links in access-control.md

GitHubPang:
In page "[Authentication, Authorisation, Access Control](https://www.rabbitmq.com/access-control.html#authorisation)",
section "Authorisation: How Permissions Work",
the links in the table were not properly formatted.

I fixed the link formatting.


acogoluegnes (approved):
Markdown syntax is not valid in HTML table, so we have to fall back to HTML syntax for the links.

acogoluegnes:
Thanks!

PR #1321, opened 2021-12-22
Provide more precise description on ha-all mode

senpl:
Previous use next line which was not really precise enough and simply not clear.


michaelklishin (approved):

michaelklishin:
Can't wait for classic mirrored queues and the whole "ha-all" section to be axed.

PR #1324, opened 2022-01-12
ra instead of raft

johanrhodin:
In the classic config format it is:
`{ra,[{wal_max_size_bytes,67108864}]}`
I assume that is the case for the advanced config format too
https://github.com/rabbitmq/rabbitmq-server/blob/aed64900f8d5d90c01d62962c91d6c0eeed269c8/deps/rabbit/apps/rabbitmq_prelaunch
/src/rabbit_prelaunch_conf.erl#L101

### DIFF
--- a/site/mqtt.md
+++ b/site/mqtt.md
@@ -225,12 +225,12 @@ In environments where the MQTT plugin is the only Raft-based feature used
 the log stored in memory will reduce memory footprint of the plugin in case of
 [high connection churn](/connections.html#high-connection-churn).
 
-The configuration key of interest is `raft.wal_max_size_bytes`:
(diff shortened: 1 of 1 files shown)


michaelklishin:
That's not correct:

``` ini
{mapping, "raft.wal_max_size_bytes", "ra.wal_max_size_bytes", [
  {datatype, integer},
  {validators, ["non_zero_positive_integer"]}
]}.

{translation, "ra.wal_max_size_bytes",
    fun(Conf) ->
        case cuttlefish:conf_get("raft.wal_max_size_bytes", Conf, undefined) of
            undefined -> cuttlefish:unset();
            Val       -> Val
        end
    end
}.
```

PR #1335, opened 2022-01-27
Change distro name for Debian Bullseye release

Brzozova:
Installation on Debian Bullseye with `buster` as a name of RabbitMQ apt repository on PackageCloud breaking the deployment and rabbitmq-server can't be loaded. I decided to deploy RabbitMQ adding to /etc/apt/sources.list.d/rabbitmq.list `bullseye`:
```
#  # Provides RabbitMQ
#
deb [signed-by=/usr/share/keyrings/io.packagecloud.rabbitmq.gpg] https://packagecloud.io/rabbitmq/rabbitmq-server/debian/ bullseye main
deb-src [signed-by=/usr/share/keyrings/io.packagecloud.rabbitmq.gpg] https://packagecloud.io/rabbitmq/rabbitmq-server/debian/ bullseye main
```
And finally, my RabbitMQ started to work. 
That would definitely save us time if this information was already available in your documentation. Thank you in advance!

### DIFF
--- a/site/install-debian.md
+++ b/site/install-debian.md
@@ -289,7 +289,7 @@ with the RabbitMQ apt repository on PackageCloud.
 | Ubuntu 20.04    | `focal`      |
 | Ubuntu 18.04    | `bionic`     |
 | Debian Buster   | `buster`     |
-| Debian Bullseye | `buster`     |
+| Debian Bullseye | `bullseye`   |
 | Debian Sid      | `buster`     |
 
 To add the apt repository to the source list directory (`/etc/apt/sources.list.d`), use:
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thanks but this is a little light on details. There is only one Debian package, so your change won't actually result in a different one being downloaded.

No too long ago PackageCloud did not support `bullseye` as a distribution name and using `buster` on Bullseye was the only option. Which is why the docs are the way they are.

It is a supported distribution now and we re-publish the same package [for it](https://packagecloud.io/app/rabbitmq/rabbitmq-server/search?dist=debian%2Fbullseye) so there is no harm in accepting this change but we do not guess on this team and what exactly "broke your deployment" is unclear.

Consider providing all details of what you tried and what you observed in the future.

michaelklishin:
This also conveniently assumes that distribution names on Cloudsmith match those on PackageCloud. That's not always the case. We have been publishing packages for [Bullseye since `3.8.15`](https://cloudsmith.io/~rabbitmq/repos/rabbitmq-server/packages/?q=distribution%3Adebian+AND+distribution%3Abullseye) so again, no harm in updating the table
but if Bullseye was out for only a couple of months this would likely not work as expected.

PR #1339, opened 2022-02-08
Current year update

venkataadithan:
Keep the year updated to ensure documentation is maintained and updted.


michaelklishin:
I'm not sure that defines whether documentation is maintained but OK.

PR #1349, opened 2022-02-17
Update tutorial-two-javascript.md

NadavK:
"RabbitMQ will *understand* that a message wasn't processed fully and will re-queue it" --> "RabbitMQ will *infer* that a message wasn't fully processed and will re-queue it"


michaelklishin:
I'm not convinced this is an improvement but if it is, please submit a single PR instead of 10 separate one liners.

PR #1389, opened 2022-04-06
Updating import on boot section of definition.md in order to mention …

DanielePalaia:
…importing from URL over https


michaelklishin (approved):

michaelklishin:
Thank you!

PR #1391, opened 2022-04-14
24.2 for 3.8.x too

johanrhodin:
If 3.9.15 is the first to support Erlang 24.3 than it can't be supported by 3.8.x.

Maybe it should say 24.2.x (as 24.2.2 is supported but not 24.3)?


johanrhodin:
oh, https://github.com/rabbitmq/rabbitmq-server/pull/4448 is backported to 3.8.29

PR #1402, opened 2022-05-03
De-emphasize dest-queue

chris-peterson:
Preserving the original exchange and routing key seems like the desirable default behavior for most cases

When using `dest-queue`, the exchange is bypassed and the routing key is overwritten.  The destructive nature of changing the original message seems to make this "advanced"

### DIFF
--- a/site/shovel-dynamic.md
+++ b/site/shovel-dynamic.md
@@ -63,7 +63,7 @@ name and a definition body which is a JSON document:
 <pre class="lang-bash">
 # my-shovel is here is the name of the shovel
 rabbitmqctl set_parameter shovel my-shovel \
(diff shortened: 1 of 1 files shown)


michaelklishin:
I'm afraid I disagree. Most Shovel users transfer queue to queue, and prefer it being an explicit configuration. We'd consider a new section that discusses how re-publishing is performed when no destination is specified but this is definitely not a good candidate for
this basic example.

michaelklishin:
Thank you for taking the time to contribute, of course :)

michaelklishin:
I personally would add a section on routing key persistence (or lack of) for shoveled messages, and cover these examples there. I'd be happy to help with that if someone put together an initial PR.

PR #1404, opened 2022-05-06
Update possible log levels

gomoripeti:
I tried to avoid ambiguous "higher severity" as it sounds like "more severe/critical" but could mean "higher severity number" ie "less severe". (Although "more critical" is not perfect either as there is a "critical" level, and "emergency" is more critical than "critical" :) )

### DIFF
--- a/site/logging.md
+++ b/site/logging.md
@@ -391,23 +391,29 @@ log.upgrade.level = none
 
 ### <a id="log-levels" class="anchor" href="#log-levels">Log Levels</a>
 
-Log levels is another way to filter and tune logging. Each log level has a severity associated with it.
-More critical messages have lower severity number, while `debug` has the highest number.
+Log levels is another way to filter and tune logging. Log levels have
+a strict ordering. Each log message has a severity from `debug` being
+the lowest severity to `critical` being the highest.
+
+Logging verbosity can be controlled on multiple layers by setting log
+levels for categories and outputs. More verbose log levels will
+include more log messages from `debug` being the most verbose to
+`none` being the least.
 
 The following log levels are used by RabbitMQ:
 
-| Log level  | Severity |
-|------------|----------|
-| `debug`    | 128      |
-| `info`     | 64       |
-| `warning`  | 16       |
-| `error`    | 8        |
-| `critical` | 4        |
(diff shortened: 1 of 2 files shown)


michaelklishin:
Do we really need to list that many?

michaelklishin (changes requested):
I think the original set of severities is perfectly sufficient for most users. They are the severities RabbitMQ actually uses as well.

gomoripeti:
thanks for the feedback, I will update the PR accordingly

is "critical" part of that reduced set? (as it was present on the logging page but not on the man page)
Also should `rabbitmqctl set_log_level` still allow all levels supported by `logger` or just the reduced list? (today it allows the extended list https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/set_log_level_command.ex#L10-L19 but prints out the reduced list https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbitmq_cli/lib/rabbitmq/cli/ctl/commands/set_log_level_command.ex#L37) this mixed behaviour also makes sense

michaelklishin:
We can keep `critical`. I can see it being used one day. I doubt anyone would need both `critical` and `emergency`, for example.

michaelklishin (changes requested):
Man pages reside in the [server repo](https://github.com/rabbitmq/rabbitmq-server/tree/v3.10.x/deps/rabbit/docs) and can be previewed with

```
man deps/rabbit/docs/rabbitmqctl.8
```

and built with either

```
bazel build "//deps/rabbit:web-manpages"
```

or

```
gmake web-manpages
```

then copied to this repo.

michaelklishin (approved):

gomoripeti:
I have pushed two new commits
- one is to disambiguate "higher severity" (I might have gone a bit too far on one of the late nights to reword this section - so please feel free to reject this commit
- other is the generated html man page from https://github.com/rabbitmq/rabbitmq-server/pull/4793 - again if rather the Core Teams does the syncing of all man pages themselves, this commit can also be dropped

thanks for your guidance and patience

PR #1405, opened 2022-05-06
Update .md files to use relative linking

dyozie:
This is a batch of changes to ensure that markdown files use relative links to other files (other markdown files and image files), instead of using the absolute path from the root of the website. 

This change makes the docs source more portable/usable for building in other contexts, like for the Tanzu RabbitMQ site. It adds no content or functionality change to the existing open source rabbitmq site.

### DIFF
--- a/site/404.md
+++ b/site/404.md
@@ -18,7 +18,7 @@ limitations under the License.
 # Page not found
 
 Oops! The page you were looking for has not been found. Please take a look at
-the [documentation ToC](/documentation.html) and [recent RabbitMQ releases](/changelog.html).
+the [documentation ToC](./documentation.html) and [recent RabbitMQ releases](./changelog.html).
 
 If you think you've found a broken link, feel free to [submit a pull request](https://github.com/rabbitmq/rabbitmq-website/)
 or report a documentation issue [on the mailing list](https://groups.google.com/forum/#!forum/rabbitmq-users).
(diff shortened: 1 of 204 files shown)


dyozie:
@pstack2021 @edbyford @ablease @michaelklishin I can't seem to add reviewers to this, so tagging you directly. 

Hopefully these changes are straightforward.  The docs were previously using a mix of relative and absolute links (with mostly absolute links). Standardizing on relative linking makes things much easier for publishing to other sites/url patterns like for the Tanzu docs.

michaelklishin:
I could not find any broken links in over 5 minutes of clicking around.

@dyozie this is a non-trivial amount of less than exciting work, thank you so much for doing it all!

michaelklishin:
Unfortunately I could fine one class of links that breaks due to CSS files hitting 404s:
any path that has path depth greater than one, e.g. `kubernetes/tanzu/installation.html`.

I reverted the deployed version to the version before this PR. You are welcome to push
any corrections to the same branch, I will pull it. Or simply submit a new PR.

dyozie:
Thanks @michaelklishin - I see that my search/replace went a little too far.  There's no need for this PR to change the css links at all, so I'll restore them.

dyozie:
Looking at my local build, the last commit seems to restore the expected styling for nested pages.

michaelklishin:
Verified and merged. It should be lived in a couple of minutes. Thanks again.

PR #1407, opened 2022-05-11
Update page.xsl for avoiding inaccessible links

yusufyilmazfr:
Hi! 👋 

I've found some errors while I was searching on RabbitMQ doc. Some links are broken, relative path was used but it is not valid in some folders so I've fixed them. 

You're able to see in this page which links are broken or inaccessible in header and footer. https://www.rabbitmq.com/tutorials/tutorial-six-java.html


michaelklishin:
References #1405.

yusufyilmazfr:
it was my pleasure. 😇 🙏

PR #1408, opened 2022-05-13
Transfer OAuth 2 examples

michaelklishin:
from https://github.com/rabbitmq/rabbitmq-oauth2-tutorial.


michaelklishin:
Note that we intentionally do not link to this guide as it is not yet ready for prime time.
It will be "soft launched" for a a few select large scale users.

PR #1418, opened 2022-06-13
Add Kubernetes to list of supported deployment methods.

MirahImage:
Kubernetes was absent from the list of supported deployment types on the homepage.


michaelklishin (approved):

michaelklishin:
Good catch.

PR #1424, opened 2022-06-20
`time_limit` of `process_data_events` is set to `None`

ChamRun:
Calling `connection.process_data_events` in loop, uses 100% CPU until response is returned, and this makes problem if server does not response.
Passing None to `process_data_events`, blocks until I/O produces actionable events.


michaelklishin (approved):

michaelklishin:
Thank you!

PR #1426, opened 2022-06-20
Removed unnecessary restore from C# tutorial one

JonasBarka:
Removed two instances of unnecessary "dotnet restore" from C# tutorial one. Restore is run by the preceding "dotnet add package".


michaelklishin:
Thank you!

PR #1428, opened 2022-06-20
Simplified recommendations on running clients

JonasBarka:
- Removed advice to close client with CTRL-C, as it closes on "enter" as printed by the client.
- Simplified advice on sending several messages, as there is not need for yet another terminal instance.


michaelklishin (approved):

michaelklishin:
Such changes can go straight into `live`.

michaelklishin:
Thank you!

michaelklishin:
Yup, and they will be deployed live a couple of minutes after merging.

PR #1432, opened 2022-06-22
Updated instructions for creating NewTask.cs

JonasBarka:
Updated instructions for creating NewTask.cs, to make it a complete description on how to get there from the old Send.cs without adding code that is part of later stages of the turtorial.


michaelklishin (approved):

michaelklishin:
Fair enough. Thank you.

PR #1439, opened 2022-06-23
Fix broken (empty) xrefs in markdown source files

dyozie:
Markdown files had a number of empty xrefs (both markdown format and html format) that were causing problems. This PR adds targets to the empty links.


dyozie:
@michaelklishin @pstack2021 these issues were breaking PDF generation on the VMware toolchain. I've already made the same changes in our release branches so this is just to cover future releases.

michaelklishin (approved):

pstack2021:
Thanks very much @dyozie

PR #1450, opened 2022-07-17
Specify the version since consumer acknowledgment timeout is enforced

b1az:
It's unclear whether "modern versions" include 3.7.x, so better specify the exact version. Which as I can tell was 3.8.0, and the change was added in [04e4d0b](https://github.com/rabbitmq/rabbitmq-server/commit/04e4d0b5ba486abb16e873b4d3df6e8792b009e6).

### DIFF
--- a/site/consumers.md
+++ b/site/consumers.md
@@ -385,7 +385,7 @@ See [.NET client guide](./dotnet-api-guide.html#basic-get) for examples.
 
 ## <a id="acknowledgement-timeout" class="anchor" href="#acknowledgement-timeout">Delivery Acknowledgement Timeout</a>
 
-In modern RabbitMQ versions, a timeout is enforced on consumer delivery acknowledgement.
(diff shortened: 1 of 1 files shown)


michaelklishin:
If 3.7 is entirely out of support, can it be modern?

michaelklishin:
Thank you for taking the time but the docs only cover the latest release series, and we try to not mention versions of something that's happened several release series ago.

b1az:
Firstly, please forget that I even mentioned the 3.7.x series. I still think it'd be more helpful to the users if they knew the exact, minor and/or patch version when a feature started to be supported. Otherwise an average, such as myself, would have to be familiar with how RabbitMQ releases are cut.

(A parallel: for the `Built-in Prometheus Support` it would've been probably unacceptable to change its documentation from `As of 3.8.0` to `In modern RabbitMQ versions`.)

PR #1454, opened 2022-07-29
Update tutorial-seven-java.md

tan-wen:
The variable name `channel` is different before and after
`channel.basicPublish(exchange, queue, properties, body);`  and `ch.waitForConfirmsOrDie(5_000);`


michaelklishin:
Thank you!

PR #1462, opened 2022-08-22
Add missing '-' to 'number' parameter

etc-crontab:
A '-' is missing before the `number` option in the example of the `rabbitmq-diagnostics log_tail` command.


michaelklishin:
Thank you but this file is generated [from `rabbitmq-diagnostics(8)`](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/docs/rabbitmq-diagnostics.8) and should not be updated directly.

michaelklishin:
I've pushed first updates in https://github.com/rabbitmq/rabbitmq-website/commit/cad4f00aad8cb9d3f1d54b8569b2a280f409e0eb but now see that this was about updating an example, not the flag argument that was also off. I'll get to that tomorrow.

PR #1464, opened 2022-08-23
Update stripped extended spec

Zerpet:
It has missing the constant for 'no-route' code 312. This constant is
present in the extended non-stripped version and it is supported by
RabbitMQ.


Zerpet:
Do we need to backport this PR to `stable` and/or `main` ?

michaelklishin:
@Zerpet yes, to keep the delta to the minimum. I will do that. We merge `live` into `stable` and `stable` into `main` every so often anyway.

PR #1467, opened 2022-08-31
Update Quorum Queues documentation to reflect Queue TTL supported since version

philleonard:
Quorum Queue Message TTL support was introduced in version `3.8.10` of `rabbitmq-server`. See:
- https://github.com/rabbitmq/rabbitmq-server/releases/tag/v3.8.10
- https://github.com/rabbitmq/rabbitmq-server/pull/2407

We discovered this when seeing errors for defining the `x-expires` parameter on a `3.8.9` cluster and coming across the above PR.

In this PR I have updated the documentation to indicate this, linking the GitHub release page (given there is no official blog post for this release). Let me know if there is a preference for this as I see not all versions are linked to release notes e.g. `3.10` in `Lazy behaviour`.

### DIFF
--- a/site/quorum-queues.md
+++ b/site/quorum-queues.md
@@ -120,7 +120,7 @@ Some features are not currently supported by quorum queues.
 | Per message persistence | per message | always |
 | Membership changes | automatic | manual  |
 | [Message TTL (Time-To-Live)](./ttl.html) | yes | yes ([since 3.10](https://blog.rabbitmq.com/posts/2022/05/rabbitmq-3.10-release-overview/)) |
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you but I'm afraid we cannot accept this. RabbitMQ 3.8 is entirely out of support. It doesn't make much sense for us to keep referencing 3.8 or 3.7 in the docs. As far as the core team is concerned, everything that's documented is available in 3.9 and/or 3.10.

Those running older versions, in particular 20+ patches behind in a no longer supported release series, are on their own.

PR #1470, opened 2022-09-05
Update Go tutorials

Zerpet:
Using `Channel.Publish` is deprecated. Tutorials should use updated API
`Channel.PublishWithContext`.


Zerpet:
I guess this should also be backported to `main` or `stable`?

PR #1472, opened 2022-09-08
Added PikaBus to the devtools list for python

hansehe:
Added the PikaBus library (https://github.com/hansehe/PikaBus) to the devtools list for python.


michaelklishin:
Thank you. We expect libraries to mature and gain traction before adding them. This project looks nice but very new.

For instance, there is only one QuickStart example and I personally don't recall any questions about this project in any community forums or chats. Could be just me, of course.

PR #1478, opened 2022-09-19
Document stream SAC and super streams

acogoluegnes:


michaelklishin:
I think that was meant for the 3.11 branch, let me revert and merge there.

michaelklishin:
Reverted #1479 in `main`, so this will go live together with other 3.11-specific docs.

acogoluegnes:
Thanks. Yes, it can go into `main`, 3.11 GA is close enough.

PR #1481, opened 2022-09-21
Updated Java tutorial RPC example to use CompletableFuture instead of…

michellammertink:
Update description as discussed in https://github.com/rabbitmq/rabbitmq-tutorials/pull/338


michaelklishin:
Thank you!

PR #1487, opened 2022-09-27
Add Oauth2 examples for multiple Idps and Rich Authorization Request

MarcialRosales:
RabbitMQ 3.11.0 supports multiple Authorization servers which are compliant with OpenID Connect.

This PR provides the following:
- New menu item 'Oauth2 plugin' in the left menu under `Plugins` item
- New page `oauth2.html`
- And update `oauth2-examples.html` page on how to configure RabbitMQ with various Idps


site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+    [<div class="Op"><code class="Fl" title="Fl">--leader-locator</code>
+    <var class="Ar" title="Ar">leader-locator</var></div>]
+    [<div class="Op"><code class="Fl" title="Fl">--initial-cluster-size</code>
+    <var class="Ar" title="Ar">initial-cluster-size</var></div>]</dt>
+  <dd>
+    <dl class="Bl-tag">
+      <dt><var class="Ar" title="Ar">super-stream</var></dt>
+      <dd>The name of the super stream to create.</dd>
+      <dt><var class="Ar" title="Ar">vhost</var></dt>
+      <dd>The name of the virtual host to create the super stream into.</dd>
```
pstack2021, line 340: For <The name of the virtual host to create the super stream into.> how about changing it to <<The name of the virtual host where the super stream is created.>

site/man/rabbitmq-streams.8.html
```diff
@@ -232,6 +232,16 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
       <var class="Ar" title="Ar">consumerinfoitem</var> can take any value from
       the list that follows:
     <dl class="Bl-tag">
+      <dt><var class="Ar" title="Ar">active</var></dt>
+      <dd>Boolean indicating whether the consumer is active or not.</dd>
```
pstack2021, line 236: Is it a boolean? I thought booleans took only true or false values? Is active = true?

site/man/rabbitmq-streams.8.html
```diff
@@ -232,6 +232,16 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+      <dt><var class="Ar" title="Ar">activity_status</var></dt>
+      <dd>Consumer activity status; one of:
```
pstack2021, line 238: Instead of "one of" How about starting a new sentence saying: The value can be one of the following:
GLOBAL comment for anywhere below where it applies

site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+    [<div class="Op"><code class="Fl" title="Fl">--initial-cluster-size</code>
+    <var class="Ar" title="Ar">initial-cluster-size</var></div>]</dt>
+  <dd>
+    <dl class="Bl-tag">
+      <dt><var class="Ar" title="Ar">super-stream</var></dt>
+      <dd>The name of the super stream to create.</dd>
+      <dt><var class="Ar" title="Ar">vhost</var></dt>
+      <dd>The name of the virtual host to create the super stream into.</dd>
+      <dt><var class="Ar" title="Ar">partitions</var></dt>
+      <dd>The number of partitions the super stream will have.</dd>
```
pstack2021, line 342: How about changing this: **The number of partitions the super stream will have.** to this: **The number of partitions in the super stream.**

site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+      <dt><var class="Ar" title="Ar">routing-keys</var></dt>
+      <dd>Comma-separated list of routing keys.</dd>
+      <dt><var class="Ar" title="Ar">max-length-bytes</var></dt>
+      <dd>The maximum size of partition streams, example values: 20gb,
```
pstack2021, line 346: Add the word "the" after the word "of" I would end the sentence there then so "The maximum size of the partition streams. For example, values: 20gb,

GLOBAL comment where ever it applies

site/management.md
```diff
@@ -285,20 +285,31 @@ rabbitmqctl set_user_tags full_access administrator
...
-to authenticate client applications and management UI users. When doing so, the management UI does
+RabbitMQ can be configured to use [JWT-encoded OAuth 2.0 access tokens](https://github.com/rabbitmq/rabbitmq-auth-backend-oauth2) to authenticate client applications and management UI users. When doing so, the management UI does
 not automatically redirect users to authenticate
 against the OAuth 2 server, this must be configured separately. Currently,
-only [UAA](https://github.com/cloudfoundry/uaa) is supported authorization server.
+RabbitMQ has been tested against the following Authorization servers:
+
+* [UAA](https://github.com/cloudfoundry/uaa)
+* [Keycloak](https://www.keycloak.org/)
+* [Oauth0](https://auth0.com/)
```
MirahImage, line 295: ~Oauth0~ -> Auth0

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
 This is true for all endpoints except `GET /definitions` and `POST /definitions`. Those
 endpoints require the token to be passed in the `token` query string parameter.
 
+### Configure Scopes to request during authentication
+
+We can configure which OAuth2 **scopes** RabbitMQ should claim when redirecting the user to the authorization server by
```
MirahImage, line 353: It is possible to configure which OAuth2 **scopes** RabbitMQ should claim when redirecting the user to the authorization server by

site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+        500mb.</dd>
+      <dt><var class="Ar" title="Ar">max-age</var></dt>
+      <dd>The maximum age of partition stream segments, using the ISO 8601
+          duration format, e.g. PT10M30S for 10 minutes 30 seconds, P5DT8H for 5
```
pstack2021, line 350: Same comment as above - Also, I would avoid use latin abbreviations such as "e.g." in documentation, use the full word here "For example: " so I would rewrite this entire sentence to: 
"The maximum age of the partition stream segments that are using the ISO 8601
          duration format. For example: PT10M30S for 10 minutes and 30 seconds or P5DT8H for 5 days and 8 hours.

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+management.oauth_scopes = "openid profile"
+</pre>
+This configuration implies that your authorization server must able to
+grant additional scopes otherwise having only these 2 scopes will not grant the user access to the management ui.
+
+### Minimum scope required and how the UI determines the username from the token
+
+RabbitMQ requires, at minimum, the `openid` scope. This is because RabbitMQ uses some claims in the *id token* to
+determine the username and display it on the top right corner of the management UI. The *id token* is sent by the
+Authorization server if we include the `openid` scope in the authorization request.
```
MirahImage, line 378: Authorization server if the `openid` scope is included in the authorization request.

site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+          days 8 hours.</dd>
+      <dt><var class="Ar" title="Ar">stream-max-segment-size-bytes</var></dt>
+      <dd>The maximum size of partition stream segments, example values: 500mb,
+          1gb.</dd>
+      <dt><var class="Ar" title="Ar">leader-locator</var></dt>
+      <dd>Leader locator strategy for partition streams. Possible values are:
+        <ul class="Bl-bullet Bl-compact">
```
pstack2021, line 357: Instead of "possible values are" should we say "The value can be one of the following:" I presume they can only use/select one so lets be clear on that

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+
+RabbitMQ reads the `user_name` claim from the *id_token*. If it does not have it, RabbitMQ uses the `sub` claim
```
MirahImage, line 380: RabbitMQ reads the `user_name` claim from the *id_token*. If the token does not contain the `user_name`, RabbitMQ uses the `sub` claim

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+### Configure OpenID Connect Discovery endpoint
+
+By default, RabbitMQ assumes the OpenID Connect Discovery endpoint is at `<management.oauth_provider_url>/.well-known/openid-configuration`. If your endpoint differs, you can set yours via `management.oauth_metadata_url` setting.
+
+RabbitMQ uses this endpoint to discover other endpoints such as **token** endpoint, **logout** endpoint, and others.
+
+### Logout workflow
+
+RabbitMQ follows the [OpenID Connect RP-Initiated Logout 1.0 ](https://openid.net/specs/openid-connect-rpinitiated-1_0.html)
+specification to implement the logout workflow. This means that the logout workflow is triggered from the Management UI when we click on the **Logout** button. Logging out from RabbitMQ management UI effectively logs the user out from
```
MirahImage, line 392: specification to implement the logout workflow. This means that the logout workflow is triggered from the Management UI when the user clicks on the **Logout** button. Logging out from RabbitMQ management UI not only logs the user out from the management UI itself but also from the Identity Provider.

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+the management UI itself but also from the Identity Provider.
+
+There are other two additional scenarios which can trigger a logout. One scenario occurs when the OAuth Token expires. Although RabbitMQ renews the token on the background before it expires, should the token expired, the user is logged out.
```
MirahImage, line 395: There are other two additional scenarios which can trigger a logout. One scenario occurs when the OAuth Token expires. Although RabbitMQ renews the token in the background before it expires, should the token expired, the user will be logged out.

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+The second scenario is when the management UI session exceeds the maximum allowed time configured on the [Login Session Timeout](https://www.rabbitmq.com/management.html#login-session-timeout).
+
+### Special attention to CSP header `connect-src`
+
+In other to support OAuth 2.0 protocol, RabbitMQ has to make asynchronous REST calls to the [OpenId Connect Discovery endpoint](#Configure-OpenID-Connect-Discovery-endpoint). If we override the default [CSP headers](#csp), we have to make sure that the `connect-src` CSP directive whitelists the [OpenId Connect Discovery endpoint](#Configure-OpenID-Connect-Discovery-endpoint).
```
MirahImage, line 400: In order to support OAuth 2.0 protocol, RabbitMQ has to make asynchronous REST calls to the [OpenId Connect Discovery endpoint](#Configure-OpenID-Connect-Discovery-endpoint). If the default [CSP headers](#csp) are overridden, it is necessary to ensure that the `connect-src` CSP directive whitelists the [OpenId Connect Discovery endpoint](#Configure-OpenID-Connect-Discovery-endpoint).

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+
+For instance, if we configured the CSP header with the value `default-src 'self'` we are, by default, setting `connect-src 'self'` which means we are denying RabbitMQ access to any external endpoint; hence disabling OAuth 2.0.
```
MirahImage, line 402: For instance, if the CSP header is configured with the value `default-src 'self'`, which by default sets `connect-src 'self'`, then RabbitMQ is denied access to any external endpoint, disabling OAuth 2.0.

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+
+In addition to the `connect-src` CSP header, RabbitMQ also needs these two CSP directives otherwise the OAuth 2.0 functionality may not work: `unsafe-eval` `unsafe-inline`.
```
MirahImage, line 404: In addition to the `connect-src` CSP header, RabbitMQ also needs the two CSP directives `unsafe-eval` and `unsafe-inline`, otherwise the OAuth 2.0 functionality may not work.

site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+    <div class="Pp"></div>
+  </dd>
+  <dt><a class="permalink" href="#list_stream_consumer_groups"><code class="Cm" title="Cm" id="list_stream_consumer_groups">list_stream_consumer_groups</code></a>
+    [<div class="Op"><code class="Fl" title="Fl">-p</code>
+    <var class="Ar" title="Ar">vhost</var></div>]
+    [<div class="Op"><var class="Ar" title="Ar">groupinfoitem
+    ...</var></div>]</dt>
+  <dd>
+    <div class="Pp"></div>
+    Lists groups of stream single active consumers for a vhost.
```
pstack2021, line 389: What does " Lists groups of stream single active consumers for a vhost." mean? 

Is it "Lists the groups of single active consumers in a stream for a vhost" if so please change to this - thanks

site/oauth2.md
```diff
@@ -0,0 +1,464 @@
...
+ * [Examples](#examples)
+
+
+## <a id="how-it-works" class="anchor" href="#how-it-works">How it works</a>
+
+### <a id="authorization-workflow" class="anchor" href="#authorization-workflow">Authorization Workflow</a>
+
+This plugin does not communicate with any OAuth 2.0 provider. It decodes an access token provided by the client and authorises a user based on the data stored in the token.
+
+The token can be any [JWT token](https://jwt.io/introduction/) which contains the `scope` and `aud` fields. The way the token was issued (such as what grant type was used) is outside of the scope of this plugin.
```
MirahImage, line 46: The token can be any [JWT token](https://jwt.io/introduction/) which contains the `scope` and `aud` fields. The way the token is issued (such as what grant type is used) is outside of the scope of this plugin.

site/oauth2.md
```diff
@@ -0,0 +1,464 @@
...
+auth_oauth2.https.peer_verification = verify_peer
+auth_oauth2.https.depth = 5
+auth_oauth2.https.fail_if_no_peer_cert = true
+auth_oauth2.https.hostname_verification = wildcard
+auth_oauth2.algorithms.1 = HS256
+auth_oauth2.algorithms.2 = RS256
+</pre>
+### <a id="resource-server-id" class="anchor" href="#resource-server-id">Resource Server ID and Scope Prefixes</a>
+
+OAuth 2.0 (and thus UAA-provided) tokens use scopes to communicate what set of permissions particular
```
MirahImage, line 205: OAuth 2.0 tokens use scopes to communicate what set of permissions a particular

site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+    <div class="Pp"></div>
+    The <var class="Ar" title="Ar">groupinfoitem</var> parameter is used to
+      indicate which group information items to include in the results. The
+      column order in the results will match the order of the parameters.
+      <var class="Ar" title="Ar">groupinfoitem</var> can take any value from the
+      list that follows:
+    <dl class="Bl-tag">
```
pstack2021, line 396: Change "can take any value from the list that follows:" to "can take any value from the
     following list:"

site/oauth2.md
```diff
@@ -0,0 +1,464 @@
...
+When RabbitMQ receives a JWT token, it validates it before accepting it.
+
+#### Must be digitally signed
+
+The token must carry a digital signature and optionally a `kid` header attribute which identifies the key RabbitMQ should
+use to validate the signature.
+
+#### Must not be expired
+
+RabbitMQ uses this field `exp` ([exp](https://tools.ietf.org/html/rfc7519#page-9)) to validate the token if present.
```
MirahImage, line 222: RabbitMQ uses the `exp` ([exp](https://tools.ietf.org/html/rfc7519#page-9)) field to validate the token, if present.

site/oauth2.md
```diff
@@ -0,0 +1,464 @@
...
+It contains the expiration time after which the JWT MUST NOT be accepted for processing.
+
+#### Audience must have/match the resource_server_id
+
+The `aud` ([Audience](https://tools.ietf.org/html/rfc7519#page-9)) identifies the recipients and/or resource_server of the JWT. By default, **RabbitMQ uses this field to validate the token** although we can disable it by setting `verify_aud` to `false`.  When it set to `true`, this attribute must either match the `resource_server_id` setting or in case of a list, it must contain the `resource_server_id`.
```
MirahImage, line 227: The `aud` ([Audience](https://tools.ietf.org/html/rfc7519#page-9)) field identifies the recipients and/or resource_server of the JWT. By default, **RabbitMQ uses this field to validate the token**, although this can be disabled by setting `verify_aud` to `false`. When set to `true`, this attribute must either match the `resource_server_id` setting or, in case of a list, it must contain the `resource_server_id`.

site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+      <dt><var class="Ar" title="Ar">consumers</var></dt>
+      <dd>Number of consumers in the group.</dd>
+      <dt><var class="Ar" title="Ar">partition_index</var></dt>
+      <dd>The stream partition index if the stream is part of a super stream, -1
+          if it is not.</dd>
+      <dt><var class="Ar" title="Ar">reference</var></dt>
```
pstack2021, line 402: How about changing this: "The stream partition index if the stream is part of a super stream, -1  if it is not." to this 

"The stream partition index if the stream is part of a super stream. The value is -1 if the stream is not part of a super stream".

site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+      <dd>The group reference (name).</dd>
+      <dt><var class="Ar" title="Ar">stream</var></dt>
+      <dd>The stream the consumers are attached to.</dd>
+    </dl>
+    <div class="Pp"></div>
+    If no <var class="Ar" title="Ar">groupinfoitem</var> are specified then
```
pstack2021, line 408: Change "are" to "is" above before the word "specified" actually I would reword the entire sentence above to: If a groupinfoitem is not specified, then....."

site/oauth2.md
```diff
@@ -0,0 +1,464 @@
...
+
+See the [wildcard matching test suite](./test/wildcard_match_SUITE.erl) and [scopes test suite](./test/scope_SUITE.erl) for more examples.
+
+Scopes should be prefixed with `resource_server_id`. For example,
+if `resource_server_id` is "my_rabbit", a scope to enable read from any vhost will
+be `my_rabbit.read:*/*`.
+
+### <a id="use-different-token-field" class="anchor" href="#use-different-token-field">Using a different token field for the Scope</a>
+
+By default the plugin will look for the `scope` key in the token, you can configure the plugin to also look in other fields using the `extra_scopes_source` setting. Values format accepted are scope as **string** or **list**
```
MirahImage, line 275: By default the plugin will look for the `scope` key in the token. The plugin can be configured to also look in other fields using the `extra_scopes_source` setting. Accepted scope value formats are **string** or **list**

site/oauth2.md
```diff
@@ -0,0 +1,464 @@
...
+ ...
+ "scope_as_string": "my_id.configure:*/* my_id.read:*/* my_id.write:*/*",
+ "scope_as_list": ["my_id.configure:*/*", "my_id.read:*/*", my_id.write:*/*"],
+ ...
+ }
+</pre>
+
+### <a id="use-tokens-with-clients" class="anchor" href="#use-tokens-with-clients">Using Tokens with Clients</a>
+
+A client must present a valid `access_token` acquired from an OAuth 2.0 provider (UAA) as the **password**
```
MirahImage, line 303: A client must present a valid `access_token` acquired from an OAuth 2.0 provider (such as UAA) as the **password**

site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+    <var class="Ar" title="Ar">reference</var>
+    [<div class="Op"><code class="Fl" title="Fl">--vhost</code>
+    <var class="Ar" title="Ar">vhost</var></div>]
+    [<div class="Op"><var class="Ar" title="Ar">consumerinfoitem
+    ...</var></div>]</dt>
+  <dd>
+    <div class="Pp"></div>
+    Lists consumers of a stream consumer group in a vhost.</dd>
+  <dt><var class="Ar" title="Ar">stream</var></dt>
+  <dd>The stream the consumers are attached to.</dd>
```
pstack2021, line 431: I would change this: "The stream the consumers are attached to" to this:

"The stream to which consumers are attached. "

site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+  <dt><var class="Ar" title="Ar">reference</var></dt>
+  <dd>The group reference (name).</dd>
+  <dt><var class="Ar" title="Ar">vhost</var></dt>
+  <dd>The virtual host of the stream.</dd>
+</dl>
+<div class="Pp"></div>
+The <var class="Ar" title="Ar">consumerinfoitem</var> parameter is used to
+  indicate which consumer information items to include in the results. The
+  column order in the results will match the order of the parameters.
```
pstack2021, line 440: remove the word "will" above, avoid future tense in documentation where possible change to "The
  column order in the results matches the order of the parameters."

site/oauth2.md
```diff
@@ -0,0 +1,464 @@
...
+in order to authenticate with RabbitMQ.
+
+To learn more about UAA/OAuth 2.0 clients see [UAA docs](https://github.com/cloudfoundry/uaa/blob/master/docs/UAA-APIs.rst#id73).
```
MirahImage, line 306: To learn more about OAuth 2.0 clients, see the [OAuth 2.0 client specification](https://www.rfc-editor.org/rfc/rfc6749#section-4.4).

site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+  <var class="Ar" title="Ar">consumerinfoitem</var> can take any value from the
+  list that follows:
```
pstack2021, line 442: Similiar comment to above - Change to "can take any value from the
 following list:
 
 GLOBAL comment, apply anywhere else this is used

site/oauth2.md
```diff
@@ -0,0 +1,464 @@
...
+    },
+    { "type" : "rabbitmq",
+      "locations": ["cluster:finance", "cluster:inventory" ],
+      "actions": ["administrator" ]
+    }
+  ]
+}
+</pre>
+
+if RabbitMQ node's `resource_server_id` is equal to `finance`, the plugin will compute the following sets of scopes:
```
MirahImage, line 449: if a RabbitMQ node's `resource_server_id` is equal to `finance`, the plugin will compute the following sets of scopes:

site/man/rabbitmq-streams.8.html
```diff
@@ -303,7 +314,155 @@ <h3 class="Ss" title="Ss" id="Stream_plugin"><a class="permalink" href="#Stream_
...
+    <ul class="Bl-bullet Bl-compact">
+      <li>active</li>
+      <li>inactive</li>
+    </ul>
+  </dd>
+  <dt><var class="Ar" title="Ar">subscription_id</var></dt>
+  <dd>The connection-scoped ID of the consumer.</dd>
 </dl>
+<div class="Pp"></div>
+If no <var class="Ar" title="Ar">consumerinfoitem</var> are specified then
```
pstack2021, line 457: Same comment as above, change to "If a consumerinfoitem is not specified, then...."

site/oauth2.md
```diff
@@ -0,0 +1,464 @@
...
+- `finance.read:primary-*/*/*`
+- `finance.write:primary-*/*/*`
+- `finance.configure:primary-*/*/*`
+- `finance.tag:administrator`
+
+## <a id="examples" class="anchor" href="#examples">Examples</a>
+
+The [RabbitMQ OAuth2 Auth Backend Examples](oauth2-examples.html) contains many example configuration files which can be used to set up several OAuth 2.0 providers such as UAA and issue tokens, which can be used to access RabbitMQ resources.
```
MirahImage, line 457: The [RabbitMQ OAuth2 Auth Backend Examples](oauth2-examples.html) contains many example configuration files which can be used to set up several OAuth 2.0 providers, including UAA, Auth0, and Azure, and issue tokens, which can be used to access RabbitMQ resources.

site/oauth2.md
```diff
@@ -0,0 +1,464 @@
...
+
+
+## License and Copyright
+
+(c) 2016-2020 VMware, Inc. or its affiliates.
```
MirahImage, line 462: (c) 2016-2022 VMware, Inc. or its affiliates.

MirahImage (changes requested):

site/management.md
```diff
@@ -285,20 +285,31 @@ rabbitmqctl set_user_tags full_access administrator
 
 ## <a id="oauth2-authentication" class="anchor" href="#oauth2-authentication">Authenticating with OAuth 2</a>
 
-RabbitMQ can be configured to use [JWT-encoded OAuth 2.0 access tokens](https://github.com/rabbitmq/rabbitmq-auth-backend-oauth2)
-to authenticate client applications and management UI users. When doing so, the management UI does
+RabbitMQ can be configured to use [JWT-encoded OAuth 2.0 access tokens](https://github.com/rabbitmq/rabbitmq-auth-backend-oauth2) to authenticate client applications and management UI users. When doing so, the management UI does
 not automatically redirect users to authenticate
```
pstack2021, line 289: I would change "When doing so" to "During this configuration, the management UI does...."

site/management.md
```diff
@@ -285,20 +285,31 @@ rabbitmqctl set_user_tags full_access administrator
...
 against the OAuth 2 server, this must be configured separately. Currently,
-only [UAA](https://github.com/cloudfoundry/uaa) is supported authorization server.
+RabbitMQ has been tested against the following Authorization servers:
```
pstack2021, line 291: Replace "has been" above with "is"....use present tense where possible in documentation

site/management.md
```diff
@@ -285,20 +285,31 @@ rabbitmqctl set_user_tags full_access administrator
...
 management.enable_uaa = true
-management.uaa_client_id = rabbit_user_client
-management.uaa_location = https://my-uaa-server-host:8443/uaa
+management.oauth_enabled = true
+management.oauth_client_id = rabbit_user_client
+management.oauth_client_secret = rabbit_user_client_secret
+management.oauth_provider_url = https://my-uaa-server-host:8443/uaa
 </pre>
 
+> IMPORTANT: Since RabbitMQ 3.10, RabbitMQ uses `authorization_code` grant type. `implicit` flow has been
```
pstack2021, line 308: Again in the Important Note, change "has been" to "is"

site/management.md
pstack2021, line 353: "We can configure...."

Should that be "You can configure"

This is a GLOBAL comment wherever there is reference to "we" if it makes sense to change it to "you" I presume it will be the user/customer doing this?

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+management.oauth_scopes = "openid profile rabbitmq.*".
+</pre>
+
+Where `rabbitmq` is the *resource_server_id* we declared in the oauth2 plugin configuration:
+<pre class="lang-ini">
+ {rabbitmq_auth_backend_oauth2, [
+    {resource_server_id, &lt;&lt;"rabbitmq"&gt;&gt;},
+</pre>
+
+In the contrary, when `management.enable_uaa = false` or when it is not set, RabbitMQ defaults to
```
pstack2021, line 367: I would remove unnecessary words in documentation - I don't think we need "In the contrary" please remove

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+<pre class="lang-ini">
+management.oauth_scopes = "openid profile"
+</pre>
+This configuration implies that your authorization server must able to
+grant additional scopes otherwise having only these 2 scopes will not grant the user access to the management ui.
+
```
pstack2021, line 373: What does this mean exactly: "This configuration implies that your authorization server must able to
grant additional scopes otherwise having only these 2 scopes will not grant the user access to the management ui."

I think the above needs to be made clearer, why are we telling them about these 2 scopes at this point in the documentation if they don't give access to the management UI, surely we should be documenting the minimum scopes that do first?

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+### Minimum scope required and how the UI determines the username from the token
+
+RabbitMQ requires, at minimum, the `openid` scope. This is because RabbitMQ uses some claims in the *id token* to
+determine the username and display it on the top right corner of the management UI. The *id token* is sent by the
+Authorization server if we include the `openid` scope in the authorization request.
+
```
pstack2021, line 379: Suggested reword for the above: 

At a minimum, RabbitMQ requires the the `openid` scope because it uses some claims in the *id token* to determine the username and to display this username on the top right corner of the management UI. The *id token* is sent by the Authorization server if it is included in the `openid` scope in the authorization request.

GLOBAL comment - avoid using "we" in documentation
Another comment - "The *id token* is sent by the Authorization server"...sent to where, please be as clear as possible in description

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+RabbitMQ reads the `user_name` claim from the *id_token*. If it does not have it, RabbitMQ uses the `sub` claim
+instead.
+
```
pstack2021, line 382: Suggested reword "If it does not have it" 

Change to:

If it is not there, RabbitMQ uses....

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+### Configure OpenID Connect Discovery endpoint
+
+By default, RabbitMQ assumes the OpenID Connect Discovery endpoint is at `<management.oauth_provider_url>/.well-known/openid-configuration`. If your endpoint differs, you can set yours via `management.oauth_metadata_url` setting.
+
```
pstack2021, line 386: Add the word "the" after "via"

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+RabbitMQ uses this endpoint to discover other endpoints such as **token** endpoint, **logout** endpoint, and others.
+
+### Logout workflow
+
+RabbitMQ follows the [OpenID Connect RP-Initiated Logout 1.0 ](https://openid.net/specs/openid-connect-rpinitiated-1_0.html)
+specification to implement the logout workflow. This means that the logout workflow is triggered from the Management UI when we click on the **Logout** button. Logging out from RabbitMQ management UI effectively logs the user out from
+the management UI itself but also from the Identity Provider.
```
pstack2021, line 393: Suggested reword: "Logging out from the RabbitMQ management UI logs the user out from
the management UI and also from the Identity Provider."

Reworded above to remove unnecessary wording - make information more concise.

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+
+There are other two additional scenarios which can trigger a logout. One scenario occurs when the OAuth Token expires. Although RabbitMQ renews the token on the background before it expires, should the token expired, the user is logged out.
+The second scenario is when the management UI session exceeds the maximum allowed time configured on the [Login Session Timeout](https://www.rabbitmq.com/management.html#login-session-timeout).
```
pstack2021, line 396: "on the background" should be "in the background" `

Change this: "should the token expired, the user is logged out." to a new sentence "if the token expires, the user is logged out.

site/management.md
pstack2021, line 400: Typo - "In other " should be "In order" actually lets leave these words out completely and reword to: "To support the OAuth 2.0 protocol, RabbitMQ makes asynchronous REST calls to the [OpenId Connect Discovery endpoint](#Configure-OpenID-Connect-Discovery-endpoint).

site/management.md
```diff
@@ -335,6 +348,61 @@ curl -i --header "authorization: Bearer &lt;token&gt;" http://localhost:15672/ap
...
+
+### Special attention to CSP header `connect-src`
+
+In other to support OAuth 2.0 protocol, RabbitMQ has to make asynchronous REST calls to the [OpenId Connect Discovery endpoint](#Configure-OpenID-Connect-Discovery-endpoint). If we override the default [CSP headers](#csp), we have to make sure that the `connect-src` CSP directive whitelists the [OpenId Connect Discovery endpoint](#Configure-OpenID-Connect-Discovery-endpoint).
+
+For instance, if we configured the CSP header with the value `default-src 'self'` we are, by default, setting `connect-src 'self'` which means we are denying RabbitMQ access to any external endpoint; hence disabling OAuth 2.0.
+
+In addition to the `connect-src` CSP header, RabbitMQ also needs these two CSP directives otherwise the OAuth 2.0 functionality may not work: `unsafe-eval` `unsafe-inline`.
+
```
pstack2021, line 405: Suggested reword above: In addition to the `connect-src` CSP header, RabbitMQ also needs the CSP directives `unsafe-eval` and `unsafe-inline`, otherwise the OAuth 2.0 functionality may not work.

site/oauth2-examples-azure.md
```diff
@@ -0,0 +1,237 @@
+# Use Azure Active Directory (Azure AD) as OAuth 2.0 server
+
+We are going to test 3 OAuth flows:
+1. Access management ui via a browser :ballot_box_with_check:
+2. Access management rest api :construction:
+3. Access AMQP protocol :construction:
+
```
pstack2021, line 7: GLOBAL comment - avoid using the word "we" how about "Let's text the following 3 OAuth flows:"

site/oauth2-examples-azure.md
```diff
@@ -0,0 +1,237 @@
...
+* Have an account in https://portal.azure.com.
+* Docker
+* Openssl
+
+## Register your app
+
+When using **Azure AD as OAuth 2.0 server**, your client app (in our case RabbitMQ) needs a way to trust the security tokens issued to it by the **Microsoft identity platform**. The first step in establishing that trust is by **registering your app** with the identity platform in Azure AD.
+
+<g-emoji class="g-emoji" alias="blue_book" fallback-src="https://github.githubassets.com/images/icons/emoji/unicode/1f4d8.png">📘</g-emoji> More details about App registration in Azure AD are available [here](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app)
+
```
pstack2021, line 19: Add a full stop at the end of the sentence above.

site/oauth2-examples-azure.md
```diff
@@ -0,0 +1,237 @@
...
+Once you have logged onto your account in [Azure Portal](https://portal.azure.com), go to **Azure Active Directory** (use the search bar if you are not able to easily find it).
```
pstack2021, line 20: Uuse numbered steps for this procedure and add full stops at the end of each step.

site/oauth2-examples-azure.md
```diff
@@ -0,0 +1,237 @@
...
+* Application (client) ID
+
+Click on the **Endpoints** tab and, on the right pane that has just opened, copy the value of **OpenID Connect metadata document** (ex: `https://login.microsoftonline.com/{TENANT_ID}/v2.0/.well-known/openid-configuration`) and open it in your browser.
+
+Note the value of the `jwks_uri` key (ex: `https://login.microsoftonline.com/{TENANT_ID}/discovery/v2.0/keys`), as you will also need it later to configure the `rabbitmq_auth_backend_oauth2` on RabbitMQ side
+
+![Azure AD JWKS URI](./img/oauth2/azure-ad-jwks-uri.png)
+
+## Create a secret for your app
+
```
pstack2021, line 49: Again use numbered steps for this procedure

site/oauth2-examples-azure.md
```diff
@@ -0,0 +1,237 @@
...
+Your application needs a **client secret** to prove its identity when requesting a token.
+
+Still on the **App registrations** page, in the left-hand menu, click on **Certificates & Secrets**, then select the **Client secrets** tab.
```
pstack2021, line 52: What if they are not still on this page, do we need to give the full sequence here on how to get to it...what if they have logged out in the meantime and now need to get back to this page

site/oauth2-examples-azure.md
```diff
@@ -0,0 +1,237 @@
...
+* **Allowed member types**: Both (Users/Groups + Applications)
+* **Value**: `Application_ID.tag:administrator` (where *Application_ID* is the value of the *Application (client) ID* noted earlier in this tutorial)
+* **Description**: briefly describe what this role aims to (here just to give admin access to the RabbitMQ Management UI)
+* **Do you want to enable this app role**: `yes` (check the box)
+
+Click on **Apply**.
+
+### Create a role to grant configure permission on all resources
+
+Click again on **Create App Role**. We are now going to create an OAuth 2.0 role that will be used to give configure access to all the resources on all the RabbitMQ vhosts.
```
pstack2021, line 86: Change "Click again" to "Click on **Create App Role** again."

site/oauth2-examples-azure.md
```diff
@@ -0,0 +1,237 @@
...
+   ]}
+ ]}
+].
+</pre>
+
+> :warning: Please update the file available in this tutorial ([here](../conf/azure/rabbitmq.config)), as it will be automatically loaded in the RabbitMQ instance that we are going to deploy later in this tutorial
+
+### Generate SSL certificate and key
+> :warning: Remember when you have registered your app on Azure AD that it only allows **https** protocol for OAuth2 **Redirect URI**? We will thus need to enable HTTPS for RabbitMQ Management UI amd its underlying API.
+
```
pstack2021, line 162: I know in VMWare documentation, the use of "Warning" in documentation is not permitted - see below - I suggest using "Important" here instead - maybe put it in caps if you want to draw more attention to it

Warning | Do not use. VMware software documentation does not require this level of notice. Warning notices are typically used in hardware documentation to alert the reader to a potentially hazardous situation that could result in serious injury or death.


-- | --

site/oauth2-examples-keycloak.md
```diff
@@ -0,0 +1,129 @@
...
+make start-perftest-producer-with-token PRODUCER=producer TOKEN=$(bin/keycloak/token producer kbOFBXI9tANgKUq8vXHLhT6YhbivgXxn)
+</pre>
+
+**NOTE**: Initializing an application with a token has one drawback: the application cannot use the connection beyond the lifespan of the token. See the next section where we demonstrate how to refresh the token.
+
+## Access AMQP protocol with Pika
+
+This section is about testing Oauth2 authentication with AMQP protocol and with Pika library. And more specifically, we
+are demonstrating how to refresh a token on a live AMQP connections.
+
```
pstack2021, line 70: Reword this: "This section is about testing Oauth2 authentication with AMQP protocol and with Pika library. And more specifically, we are demonstrating how to refresh a token on a live AMQP connections.

TO THIS:

In the following information, Oauth2 authentication is tested with the AMQP protocol and the Pika library. These tests specifically demonstrate how to refresh a token on a live AMQP connection.

site/oauth2-examples-oauth0.md
```diff
@@ -0,0 +1,134 @@
+# Use https://auth0.com/ as OAuth 2.0 server
+
+We are going to test 3 OAuth flows:
+1. Access management ui via a browser
+2. Access management rest api
+3. Access AMQP protocol
+
```
pstack2021, line 7: Same as previous comments above, avoid using "We" in documentation - For the above, use "Let's test 3 OAuth flows:"

site/oauth2-examples-oauth0.md
```diff
@@ -0,0 +1,134 @@
...
+## Prerequisites to follow this guide
+
+* Have an account in https://auth0.com/.
+* Docker
+
+## Create rabbitmq API
+
+In OAuth0, resources are mapped to Application APIs. Once you have logged onto your account in https://auth0.com/, go to dashboard > Applications > APIs > Create an API.
+
```
pstack2021, line 16: Should the UI sequence "dashboard > Applications > APIs > Create an API" be bolded to be consistent with all other UI paths referenced in this doc

site/oauth2-examples-oauth0.md
```diff
@@ -0,0 +1,134 @@
...
+* Give it the name `rabbitmq`. The important thing here is the `identifier` which must have the name of the *resource_server_id* we configured in RabbitMQ. This `identifier` goes into the `audience` JWT field. In our case, it is called `rabbitmq`
+* Choose `RS256` as the signing algorithm
+* Enable **RBAC**
+* Enable **Add Permissions in the Access Token**
+
```
pstack2021, line 21: GLOBAL - same as previous comments:
Use numbered steps for procedures
Add full stops at the end of sentences
etc

site/oauth2-examples-oauth0.md
```diff
@@ -0,0 +1,134 @@
...
+* `rabbitmq.configure:*/*`
+* `rabbitmq.tag:administrator`
+
+### Create an OAuth client for the Management UI
+
+By default, for every API we create, an *Application* gets created using the API's `identifier` as its name.
+An *Application* requests an **OAuth client**.
+
+Go to dashboard > Applications, and you should see your application listed. An application gives us a *client_id*, a *client_secret* and a http endpoint called *Domain* where to claim a token.
+
```
pstack2021, line 37: Again should **dashboard > Applications** be bolded?

site/oauth2-examples-oauth0.md
```diff
@@ -0,0 +1,134 @@
...
+## Create Application rabbitmq-management
+
+An application gives us the client-id and client-secret for the management ui to authenticate on behalf
+of the end user.
```
pstack2021, line 41: "ui" should be in uppercase "UI"

site/oauth2-examples-oauth0.md
```diff
@@ -0,0 +1,134 @@
...
+
+In setting we choose:
```
pstack2021, line 43: Should this be "In the settings, choose"?

site/oauth2-examples-oauth0.md
```diff
@@ -0,0 +1,134 @@
...
+
+
+## Create a new user
+
+You can use your current OAuth0 user to login to RabbitMq or create a dedicated user for that. Up to you.
+
+### Authorize rabbitmq-management application
+
+Go to "Authorized Applications", click on "Authorize" and select all the scopes.
+
```
pstack2021, line 58: Again should the UI options to select here be bolded, they are bolded in other procedures: "Go to "Authorized Applications", click on "Authorize" and select all the scopes. "

Keep styling consistent in documentation where possible

GLOBAL comment above and below where applicable.

site/oauth2-examples-oauth0.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Configure RabbitMQ with OAuth0 signing key
+
+From Oauth0 dashboard, go to Settings > List of Valid Keys, and "Copy Signing Certificate" from the **CURRENTLY USED** signing key.
+
+Create `/tmp/certiicate.pem` and paste the certificate.
+
+Then run `openssl x509 -in /tmp/certificate.pem -pubkey -noout > /tmp/public.pem` to extract the public key from the certificate. And paste the public key into `rabbitmq.config`.
+
```
pstack2021, line 73: Can't start a sentence with "And" just run it into the previous sentence so: "Run `openssl x509 -in /tmp/certificate.pem -pubkey -noout > /tmp/public.pem` to extract the public key from the certificate and paste the public key into `rabbitmq.config`.

Also there is no need for the word "Then" at the start of the sentence above, hence I have removed it in the rewrite.

site/streams.md
```diff
@@ -224,6 +224,46 @@ but some have some queue specific behaviour.
  * [Consumer acknowledgements](./confirms.html) (keep [QoS Prefetch Limitations](#global-qos) in mind)
  * Cancellation of consumers
 
+### <a id="single-active-consumer" class="anchor" href="#single-active-consumer">Single Active Consumer</a>
+
+Single active consumer for streams is a feature available in RabbitMQ 3.11 and more.
+It provides _exclusive consumption_ and _consumption continuity_ on a stream.
```
pstack2021, line 230: "Single active consumer for streams is a feature available in RabbitMQ 3.11 and more."`

What does "and more" mean above??? later release to 3.11? If its in 3.11, it will be in later releases by default? Delete "and more" or else clarify further -thanks

site/streams.md
```diff
@@ -224,6 +224,46 @@ but some have some queue specific behaviour.
...
+* Consumption continuity is maintained: a consumer from the group will take over if the active one stops or crashes.
+
+A [blog post](https://blog.rabbitmq.com/posts/2022/07/rabbitmq-3-11-feature-preview-single-active-consumer-for-streams/) provides more details on single active consumer for streams.
+
+### <a id="super-streams" class="anchor" href="#super-streams">Super Streams</a>
+
+Super streams are a way to scale out by partitioning a large stream into smaller streams.
+They integrate with [single active consumer](#single-active-consumer) to preserve message order within a partition.
+Super streams are available in RabbitMQ 3.11 and more.
+
```
pstack2021, line 246: Again same comment as above what is the reference to "and more"?? Delete or explain further - thanks

site/streams.md
```diff
@@ -224,6 +224,46 @@ but some have some queue specific behaviour.
...
+A super stream is a logical stream made of individual, regular streams.
+It is a way to scale out publishing and consuming with RabbitMQ Streams: a large logical stream is divided into partition streams, splitting up the storage and the traffic on several cluster nodes.
+
+A super stream remains a logical entity: applications see it as one “big” stream, thanks to the smartness of client libraries.
+The topology of a super stream is based on the [AMQP 0.9.1 model](./tutorials/amqp-concepts.html), that is exchange, queues, and bindings between them.
+
+Even though it is possible to create the topology of a super stream with any AMQP 0.9.1 library or with the [management plugin](./management.html), but the `rabbitmq-streams add_super_stream` command is a handy shortcut.
+Here is how to create an `invoices` super stream with 3 partitions:
```
pstack2021, line 254: The word "but" above it out of context, how about: 

"You can create the topology of a super stream with any AMQP 0.9.1 library or with the [management plugin](./management.html) or you can also create it by running this command: `rabbitmq-streams add_super_stream`.

pstack2021 (commented):
@MarcialRosales here are my comments - thanks

site/oauth2-examples-azure.md
MarcialRosales: I'd say it is totally fine to remind them on which page they should be now .. by saying "Still on the **App registration** page" ...

site/oauth2-examples-oauth0.md
```diff
@@ -12,21 +13,23 @@ Let's test the following 3 OAuth flows:
 
 ## Create RabbitMQ API
 
-In OAuth0, resources are mapped to Application APIs. Once you have logged onto your account in https://auth0.com/, go to **dashboard > Applications > APIs > Create an API**.
+In OAuth0, resources are mapped to Application APIs.
```
MirahImage, line 16: ~OAuth0~ -> Auth0

MirahImage (changes requested):

site/streams.md
MarcialRosales: This is related to streams not part of this PR about Oauth 2.0

site/streams.md
MarcialRosales: This is related to streams not part of this PR about Oauth 2.0

MirahImage (approved):

PR #1488, opened 2022-09-27
Update minimum required Erlang version matrix

gomoripeti:
This is based on what is enforced by `rabbit_prelaunch_erlang_compat.erl`.
(I did not modified unsupported 3.8.x versions, though OTP 21.3 was allowed up to 3.8.13)


gomoripeti:
already mostly fixed on live

PR #1490, opened 2022-09-29
Correct that QQ support customer priority

dmorneau:
Per 3.8.10 release notes:

> [Quorum queues](https://www.rabbitmq.com/quorum-queues.html) now support [consumer priority](https://www.rabbitmq.com/consumers.html#priority).
> GitHub issue: https://github.com/rabbitmq/rabbitmq-server/pull/2451


dmorneau:
@pivotal-cla This is an obvious fix

PR #1491, opened 2022-10-02
Erlang support matrix changes

carlhoerberg:
3.9.{22,23} supports erlang 25 according to their release notes. And rabbitmq 3.10.0-4 doesn't support erlang higher than 24.3 according to to debian packages Depends which sets a hard max limits on the erlang version.


michaelklishin:
Thank you. We do not plan to list 3.9.x as Erlang 25-compatible. Both because we do not test it on 25 all that much, and because we want people who run on 25 to move to 3.10.

carlhoerberg:
ok, but what about 3.10.0-4? can't install them on debian with erlang 25 and higher. still list as supported?

PR #1503, opened 2022-10-19
Renames amqp.node to amqplib

adamtajti:
It looks like that the original link for `amqp.node` redirects to `amqplib`. I was a bit confused at first, cause I didn't see `amqplib` mentioned on this page, even though the tutorials used it.


michaelklishin (approved):

michaelklishin:
Thank you, indeed it looks like the repo has been transferred from Michael B's personal account to an org.

We may have to update our tutorials in this repo as well as https://github.com/rabbitmq/rabbitmq-tutorials/ then.

PR #1506, opened 2022-10-21
Updating the redirecting links

sjyothi54:
Hey Team
There is a problem with this redirecting links so I am updating this.
We’d love to provide a smooth experience for RabbitMQ users who need to use monitoring solutions, such as New Relic.
Thank you for your consideration.


sjyothi54:
Thank you

PR #1507, opened 2022-10-21
add use cases section to DLX, with the example of NServiceBus

adamralph:


michaelklishin:
Thanks. I'm afraid we don't want to promote this unintentional use of DLX.

Not sure what links to a specific product would give to the reader, who is very likely a user of OSS RabbitMQ.

PR #1508, opened 2022-10-21
Update rabbitmq-diagnostics.8.html

pstack2021:
Changed "disabled" to "deactivated"


michaelklishin:
This page is produced from `man` pages in the server repo. I will update them and regenerate.

PR #1517, opened 2022-10-24
Update vhosts.md

pstack2021:


michaelklishin:
This is a good example of when "deactivate" does not work. I've replaced it with "stop", "block", and similar verbs in a couple of other guides.

PR #1540, opened 2022-10-25
Replacing references of "kill" in documentation

pstack2021:


michaelklishin (approved):

michaelklishin:
I like these changes! We can also use "stop" as a synonym for "terminate" when talking about apps or nodes or operating system processes.

PR #1554, opened 2022-11-07
docs: update tutoria-one-java.md

### LINKED ISSUE #1553: docs: update tutoria-one-java.md
Hi, guys :. When I read [tutorial-one-java](https://www.rabbitmq.com/tutorials/tutorial-one-java.html), origin documentation mentioned this：
> Note we can use a try-with-resources statement because both Connection and Channel implement `java.io.Closeable`. This way we don't need to close them explicitly in our code. 

 but actually I think Connection and Channel implement `java.lang.AutoCloseable`, so they can use a try-with-resources statement 🤔 .

zhicheng-ning:
- change `java.io.Closeable` to `java.lang.AutoCloseable`

close #1553

### DIFF
--- a/site/tutorials/tutorial-one-java.md
+++ b/site/tutorials/tutorial-one-java.md
@@ -110,7 +110,7 @@ machine we'd simply specify its hostname or IP address here.
 
 Next we create a channel, which is where most of the API for getting
 things done resides. Note we can use a try-with-resources statement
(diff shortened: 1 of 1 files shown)


michaelklishin (approved):

michaelklishin:
Good catch. Thank you!

PR #1564, opened 2022-11-29
update rabbitmqctl to use inclusive info keys

illotum:
Amended several mentions of `slave_pids`, updated build instructions to refer to bazel

Regenerated rabbitmqctl man-page. Opted to do it only for one page since my `mandoc` apparently produces a significantly different result. Subject to discussion.

### DIFF
--- a/site/ha.md
+++ b/site/ha.md
@@ -240,7 +240,7 @@ It is possible to list queue leader and mirrors using `rabbitmqctl list_queues`.
 example we also display queue policy since it's highly relevant:
 
 <pre class="lang-bash">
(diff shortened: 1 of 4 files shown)


michaelklishin:
Now that `3.11.4` is out, we can do this. It might cause some confusion but a comment in doc examples would do.

michaelklishin:
For some reason the web man page generator now produces `class=00` which our LXML parser does not like. I have enquoted the value manually, I guess we'll have to look into a way of solving this in the Bazel task.

CC @pjk25

PR #1584, opened 2023-01-03
Add docs for prometheus detailed metrics endpoint

iliakur:
Addresses #1579. I'm not sure this requires a contributor agreement, since I'm essentially copy/pasting from a different rabbitmq repo.


michaelklishin:
Merged with some edits, thank you!

michaelklishin:
@pivotal-cla obvious fix

PR #1587, opened 2023-01-16
Added a hint for the key import commands

shekeriev:
Because of the hash algorithm in use (SHA1) for the Erlang and RabbitMQ server repository, the appropriate rpm import commands will fail. A possible workaround is to switch the crypto policies to legacy.


michaelklishin:
Thank you!

It's been a while since we've rolled our signing key, maybe we should do it again in 2023 and not use SHA-1 this time around :)

PR #1588, opened 2023-01-19
add note about some frameworks providing tuning of concurrency

adamralph:


michaelklishin:
Thank you but this seems a bit redundant. The previous sentence says

> The pool usually has controllable degree of concurrency

We do not usually link to individual 3rd party projects in guides such as Consumers.

PR #1591, opened 2023-01-25
Expand documentation for `/metrics/detailed` endpoint

iliakur:
I noticed there were some details missing on how to construct queries to the detailed endpoint.

These docs are almost verbatim copy/paste from the plugin `README` in the server repo, I will remove those in a follow-up PR.


michaelklishin:
Thank you!

PR #1607, opened 2023-02-20
Added Lepus (scala library) to devtools page

hnaderi:


michaelklishin:
We usually do not put newer projects first on the list but support for Scala JS and Scala native seems to be fairly unique to this client, so fine.

Issue #1608, opened 2023-02-24
Default Virtual Host and User section in access control doc lacks info for the case when definitions are imported in a blank node 


databu:
I'm using the RabbitMQ docker image and the default guest user for local testing with docker compose. However, after upgrading from some older version to 3.11 the default user and password failed to auth with an invalid credentials error.

I followed everything in the [access control doc ](https://www.rabbitmq.com/access-control.html) and other hints on the web, up to manually setting `default_user` and `default_pass` in `rabbitmq.conf`, but nothing worked.

Until I found this unimposing remark buried somewhere in the text in https://www.rabbitmq.com/definitions.html :

> The definitions in the file will not overwrite anything already in the broker. However, if a blank (uninitialised) node imports a definition file, it will not create the default virtual host and user.

We do import a definitions file to pre-define some queues, and so this was the reason. I've now fixed it by re-defining the default user and their permissions in our definitions file. 

My request is to add a remark about this in the [Default Virtual Host and User](https://www.rabbitmq.com/access-control.html#default-state) section in the access control doc.

michaelklishin:
It's a fairly large doc site, we cannot put every "unimposing remark" into every guide.

RabbitMQ's core team suggestion to anyone who can to stop using the default user in production, leave alone
create a user with the default credentials when importing definitions.

databu:
I called it "unimposing" as a criticism. 

As I mentioned, we're using it only in testing. But independent of that, it took me days to just find out what the problem was.

PR #1609, opened 2023-02-24
Add remark about no default user created when importing definitions i…

databu:
…n blank node

### DIFF
--- a/site/access-control.md
+++ b/site/access-control.md
@@ -87,6 +87,24 @@ It is advisable to [pre-configure a new user with a generated username and passw
 the `guest` user or at least [change its password](rabbitmqctl.8.html#change_password)
 to reasonably secure generated value that won't be known to the public.
 
+Also note that if a blank (uninitialised) node [imports a definition file](https://www.rabbitmq.com/definitions.html#import-on-boot), it will not create the default virtual host and user. An equivalent default user can be created by including these defintions in the imported definitions file:
+```
+    "users": [
+        {
+            "name": "guest",
+            "password": "guest",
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you but we would not recommend anyone who imports definitions to use the default user.

On top of that, the same fact is already mentioned [in the Definitions guide](https://rabbitmq.com/definitions.html#import-on-boot).

databu:
As I mentioned, we're using this only in testing. It took me several days of debugging and research until I found this notice in the definitions doc. Please consider accepting the PR when I either:
1. add your remark that using a default user with imported definitions is not recommended, or
2. removing the guest user example.

michaelklishin:
@databu from the [Access Control](https://rabbitmq.com/access-control.html#default-state) guide:

> It is advisable to [pre-configure a new user with a generated username and password](https://rabbitmq.com/access-control.html#seeding) or [delete](https://rabbitmq.com/rabbitmqctl.8.html#delete_user) the guest user or at least [change its password](https://rabbitmq.com/rabbitmqctl.8.html#change_password) to reasonably secure generated value that won't be known to the public.

I'm not sure what example you are referring to.

databu:
The example that I had added in my PR. Again, I'm talking about a test system and the fact that it took me days to find out the issue. Because I was looking at the access control doc, which has no indication that importing definitions doesn't create the default user. This was also different in older releases. I didn't think of looking in the definitions doc because that just didn't occur to me. 
Shall we continue this discussion in [the issue](https://github.com/rabbitmq/rabbitmq-website/issues/1608) instead?

PR #1614, opened 2023-02-28
Introduce preferred_username_claims setting in cuttlefish config style

MarcialRosales:


michaelklishin:
This should be backported to `stable` because it documents something that's yet to ship in `3.11.10`.

PR #1616, opened 2023-03-01
Add section for topic exchanges in Oauth2.md

MarcialRosales:
Very recently there has been a number of queries/issues reported around using topic exchanges in the context of Oauth2. 
This PR adds a section that explains what scopes are required to operate on a topic exchange. This information was already in the docs but clearly it was not very obvious to most of the users.

### DIFF
--- a/site/oauth2.md
+++ b/site/oauth2.md
@@ -29,6 +29,7 @@ This guide covers
     * [Resource Server ID and Scope Prefixes](#resource-server-id)
     * [Token validation](#token-validation)
     * [Scope-to-Permission Translation](#scope-translation)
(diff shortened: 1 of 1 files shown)


michaelklishin:
@MarcialRosales can we please separate `man` page regeneration from the rest of the changes, or are they essential for documenting topic permissions here? The diff is massive, so I cannot tell.

MarcialRosales:
@michaelklishin at first I based my single commit on `main` branch but then I decided that this change should really go to `live` and when I rebased it , i must have brought these unwanted 2 extra commits. 
I have cleared the commits. Thanks for reviewing it

michaelklishin (approved):

PR #1618, opened 2023-03-03
Add new section in which-erlang.md for version 3.9.28

nordlundj:
RabbitMQ version 3.9.28 requires Erlang 24.3 as minimum version

At the moment, the which-erlang.md states 24.2 as minimum version while the deb package and release notes states that it requires 24.3. 
Added a new section for RabbitMQ version 3.9.28 with notes copied from the 3.10 versions with the same Erlang requirements.

I am however, unsure of the maximum version that 3.9.28 supports.

### DIFF
--- a/site/which-erlang.md
+++ b/site/which-erlang.md
@@ -329,6 +329,38 @@ source, including specific tags from GitHub, a much more pleasant experience.
     <td>
       <ul>
         <li>3.9.28</li>
+      </ul>
+    </td>
+    <td>
+      <ul>
+        <li>24.3</li>
+      </ul>
+    </td>
+    <td>
+      <ul>
+        <li>25.2</li>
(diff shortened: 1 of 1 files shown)


michaelklishin:
RabbitMQ 3.9.28 itself [does require 24.2](https://github.com/rabbitmq/rabbitmq-server/blob/v3.9.x/deps/rabbit/apps/rabbitmq_prelaunch/src/rabbit_prelaunch_erlang_compat.erl#L9).

`24.3.2` is listed in the [Debian package control file](https://github.com/rabbitmq/rabbitmq-packaging/blob/v3.9.x/debs/Debian/debian/control).

I'm fine with accepting this since 24.2.x no longer gets any patch releases.

michaelklishin:

PR #1621, opened 2023-03-09
Document set_permissions_globally CLI command

Zerpet:
## Summary

This command was introduced in RMQ 3.11.9 + 3.10.18 via https://github.com/rabbitmq/rabbitmq-server/pull/7197 and its backports. Documentation on the website was missing. The CLI already includes a help page documenting this command.


michaelklishin:
These guides are produced from [man pages](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/docs/rabbitmqctl.8).

Zerpet:
🥲

PR #1629, opened 2023-03-24
3.12: add a classic-queues page to the documentation

lhoguin:
It has updated information about classic queues and aims
to replace persistence-conf and lazy-queues ultimately.

* [x] It has TODOs
* [x] It probably should be linked to
* [x] It should be reviewed from a technical POV
* [x] It should be reviewed from a grammar POV
* [ ] persistent-conf.html should redirect to classic-queues.html and be removed

### DIFF
--- a/site/admin-guide.md
+++ b/site/admin-guide.md
@@ -149,7 +149,7 @@ see <a href="./documentation.html">All Documentation Guides</a>.
  * [Reliable Message Delivery](reliability.html)
 
 
-## Message Store and Resource Management
+## Resource Management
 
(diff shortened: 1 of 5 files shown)


lhoguin:
@michaelklishin Please see the OP for the details on the single redirect that will be necessary for the time being. Thanks!

michaelklishin (approved):

lhoguin:
I think the live website doesn't have the redirect yet, if I go to https://rabbitmq.com/persistent-conf.html I just get an error.

PR #1636, opened 2023-04-05
Add note to AzureAD documentation about variability of role ID prefix

kratkyzobak:
From current documentation it seems, that ClientID as AppRole prefix is mandatory. 

Problem is, this can lead to headcaches for anyone who tries to create AAD application by Terraform / ARM / any manifest declaration outside of portal. You have to be able to declare your app with all roles before you know Client ID.

Solution is well documented, but this "step-by-step" tutorial is hiding it.

### DIFF
--- a/site/oauth2-examples-azure.md
+++ b/site/oauth2-examples-azure.md
@@ -75,6 +75,8 @@ App roles are defined by using the [Azure portal](https://portal.azure.com) duri
     * **Value**: `Application_ID.tag:administrator` (where *Application_ID* is the value of the *Application (client) ID* noted earlier in this tutorial)
     * **Description**: briefly describe what this role aims to (here just to give admin access to the RabbitMQ Management UI)
     * **Do you want to enable this app role**: `yes` (check the box)
+  
+    <g-emoji class="g-emoji" alias="bulb" fallback-src="https://github.githubassets.com/images/icons/emoji/unicode/1f4a1.png">💡</g-emoji> You can use any other predefined constant instead of `Application_ID`. See full documentation [here](https://github.com/rabbitmq/rabbitmq-oauth2-tutorial#about-permissions).
 
 4. Click on **Apply**.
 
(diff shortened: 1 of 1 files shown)


kratkyzobak:
Oh.. this is not documentation problem and it's really causing headcaches :/

https://github.com/rabbitmq/rabbitmq-server/issues/7861

MarcialRosales:
@kratkyzobak I think the appRole should have the `application_id ` which is equivalent to `resource_server_id` in RabbitMQ if I am not wrong, not the `ClientId`. Do you still think that note is required? I did not build personally the tutorial on how to integrate with Azure. It was contributed by the community but I did follow the tutorial and it did work...

kratkyzobak:
@MarcialRosales based on my investiagtion made in https://github.com/rabbitmq/rabbitmq-server/issues/7861:

- If you want to use only one application (as of website tutorial), you need to use `resource_server_id` equaling to Client ID as `resource_server_id` is then promoted into OAuth URL into `resource` param. If you use `application uri` instead of Client ID you get `AADSTS90009 - TokenForItselfMissingIdenticalAppIdentifier - The application is requesting a token for itself. This scenario is supported only if the resource that's specified is using the GUID-based application ID.`
- If you use two applications, you use one for SPA flow (as in documentation) and second one for app roles definition. In this case you CAN use `application uri` of second application instead of client id in `resource_server_id` and it will work (in combination with issuing version 1 token to enable audience checking).

This PR is wrong (proposed text is not true). Current documentation is good for click-ops scenario where you want primairly Management UI behind SSO. 

I can probably provide some better explanation of terraform code which I provided in my comment in https://github.com/rabbitmq/rabbitmq-server/issues/7861 as new documentation, where I would go by steps "Create application to assign roles" and then (as optional step) "Create application for Management UI OAuth access". Don't know in what timespan I would be able to do it. If it's OK to make documentation "Create AppRegistrations for RabbitMQ OAuth2 plugin using terraform" it would be probably quickly doable by me. Documentaiton like current with Azure Portal screenshots would take me more time.

michaelklishin:
Closing because the above comment concludes that we need a different explanation/suggestion.

PR #1640, opened 2023-04-25
Mention new ways of setting Consumer Timeout

ikvmw:
see  #5437

### DIFF
--- a/site/consumers.md
+++ b/site/consumers.md
@@ -419,6 +419,35 @@ The timeout can be deactivated using [`advanced.config`](configure.html#advanced
 ].
 </pre>
 
+In the recent RabbitMQ versions the timeout value can also be set on per-queue and per-consumer basis as well as queue policy:
+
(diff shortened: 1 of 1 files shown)


michaelklishin (changes requested):
We cannot provide an Erlang consumer example and assume that would be enough. We can provide Java and .NET examples, and perhaps a Python one, and leave it at that.

johanrhodin (changes requested):
`In the recent RabbitMQ versions` -> `In recent RabbitMQ versions`

BUT I'd prefer if it said more explicitly which version(s) (`3.10.x`, `3.11.y`) it was introduced in since a 9 months from now we don't know what "recent" was in April 2023.

michaelklishin:
@johanrhodin that's what change logs are for. We usually remove specific version mentions in a year or so.

michaelklishin (commented):
This fails to render because `<` and `>` must be escaped.

PR #1665, opened 2023-06-05
add easypanel to docs

ivanciucdragos1:


michaelklishin:
Thank you for taking the time. We will not accept this PR.

Easypanel is a commercial product, and I doubt it is currently as widely used as Debian or RPM packages, Kubernetes, Homebrew, or 2-3 of the most popular RabbitMQ aaS providers out there.

This looks like a play to get free marketing. You even placed it right above the Kubernetes Operator guide. Yeah, right.

PR #1670, opened 2023-06-10
Update clustering.md

mavaddat:
The language _recommended against_ can be read as an affirmative endorsement when skimming the text. A less ambiguous and more vernacular English formulation is to ***discourage*** when _recommending against_ an action.


michaelklishin (approved):

michaelklishin:
Thank you!

PR #1683, opened 2023-07-10
add ha-* arguments to operator policies documentation

stefanmoser:
`ha-mode`, `ha-sync-mode`, and `ha-params` were added as valid argument in operator policies. This PR updates the documentation to reflect those changes.

* https://github.com/rabbitmq/rabbitmq-server/pull/7368
* https://github.com/rabbitmq/rabbitmq-server/pull/8043

Now that not all arguments are numeric, I decided the use a table to describe the conflict resolution for each argument. Please provide feedback if this makes it easy for the reader to understand how conflict resolution works.

Question: Do the docs normally state if something is new or what version support starts in? Or just treat everything as it applies to the latest version? Support for `ha-*` in operator policies was introduced in 3.11.15.

### DIFF
--- a/site/parameters.md
+++ b/site/parameters.md
@@ -401,14 +401,14 @@ semantics, they are limited only to a few arguments:
  * message-ttl
  * max-length
  * max-length-bytes
-
-The arguments above are all numeric. The reason for that is explained
-in the following section.
+ * ha-mode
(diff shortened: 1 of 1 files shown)


michaelklishin:
Merged with a small clarification. Thank you!

PR #1684, opened 2023-07-10
document disabling modification of operator policies in HTTP API and Web UI

stefanmoser:
Document how modification of operator policies via the HTTP API and Web UI can be disabled with configuration.

* https://github.com/rabbitmq/rabbitmq-server/pull/7165


michaelklishin:
I've added a dedicated doc section header and corrected code formatting. Thank you.

PR #1695, opened 2023-07-26
Update tutorial-two-python.md to use python f strings

tatzati:


michaelklishin:

PR #1697, opened 2023-07-27
docs: correct link added for Management API

grinish21:
- The older link was returning a 404 for the HTTP API reference.


michaelklishin:
Good catch, thank you.

PR #1702, opened 2023-08-09
Black h1-h4 and use 1.25 ration for sizes. Also fix STOMP doc structure

deadtrickster:


michaelklishin (approved):

michaelklishin:
This is definitely much better!

PR #1706, opened 2023-08-14
which-erlang.md: typo: RabbitMQ 3.12, not Erlang 3.12

aburch:


michaelklishin:
@pivotal-cla obvious fix

PR #1707, opened 2023-08-15
Quorum Queues guide: improve the section on length limit

maestroes:
Following the [discussion](https://groups.google.com/g/rabbitmq-users/c/2eo3z1eZBiY)

### DIFF
--- a/site/quorum-queues.md
+++ b/site/quorum-queues.md
@@ -172,6 +172,11 @@ The `drop-head` and `reject-publish` overflow behaviours are supported but they
 do not support `reject-publish-dlx` configurations as Quorum queues take a different
 implementation approach than classic queues.
 
+The current implementation of `reject-publish` overflow behaviour does not strictly
+enforce the limit and allows a quorum queue to overshoot its limit by at least
+one message, therefore it should be taken with care in scenarios where a precise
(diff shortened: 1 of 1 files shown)


kjnilsson:
Hi thanks!

The section you added doesn't document the fact that the limit is triggered 1 message before what is configured. Also the fact that is can overshoot is mentioned in the following section. Would you mind revising?

maestroes:
Hi Karl, correct me please if I got it wrong.

> the limit is triggered 1 message before what is configured

Did you mean `the limit is triggered 1 message after what is configured`?

My observation was that `x-max-length: N` setting allows to publish `N+1` messages and then starts to reject new messages. I've tried to describe this behaviour by `allows a quorum queue to overshoot its limit by at least one message` phrase. Should it be more specific?

AFAIU the following section that already contains the caution on possible overshoot is limited to the specific case of in-flight messages on channels that have not yet been notified of the limit reached. Should it be rephrased as a specific subcase of the general overshoot possibility described in a new section?

Thanks

michaelklishin (approved):

michaelklishin:
The wording about the somewhat imprecise nature of the limit sounds good to me, that's the point, really.

PR #1709, opened 2023-08-15
Fix broken rpm html link

zionio:
Fix the broken link in https://www.rabbitmq.com/news.html 

`<a href="/intall-rpm.html">dnf (yum) repositories</a>` -->  https://www.rabbitmq.com/intall-rpm.html  ❌ 

to

`<a href="/install-rpm.html">dnf (yum) repositories</a>` -->  https://www.rabbitmq.com/install-rpm.html  ✅ 


michaelklishin:
Thank you!

PR #1711, opened 2023-08-17
SEO updates in Queue topics - The approach I am taking

pstack2021:
@michaelklishin and @dyozie can you please review this 1st PR for SEO updates in the OSS content just so you can see the approach that I am taking to these updates and the logic behind it. 

If you both agree with the approach etc, I will continue along this route for the other topics I need to update. @michaelklishin I will include you in all PR reviews, it is optional whether you want to review or not (perhaps a quick scan would be great), @dyozie, I will include you in the review of the first PR to get your feedback - many thanks.

SEO updates in Queues topic:

I approached it this way:
 - Used the product-name; constant in the H1 title
 - Used product-name; constant in the next H2 title (it is the most important heading in this topic - What is a RabbitMQ Queue?)
 - Then used product-name; constant in the short description.

And that is it, I have not used product-name; constant anymore in this topic, I don't think we need to, yes there will be instances in the topic of both the commercial product name and the OSS "RabbitMQ" product name but that is ok, I think users will understand that and to be honest I don't think they will even think twice about it if they are still in the topic and reading through it. The main goals we are trying to achieve here from an SEO perspective are: a) Clear, crisp titles, and short description particularly at the beginning of the topic and b) when this OSS topic is pulled into the different commercial doc sets (K8/OCI/OVA) we can quickly identify from the start of the topic what commercial RabbitMQ offering we're referring to (product-name; constant allowing us to do this (if it works) because it will be replaced by the specific product name when pulled into the commercial doc set)

### DIFF
--- a/site/queues.md
+++ b/site/queues.md
@@ -19,7 +19,7 @@ limitations under the License.
 
 ## What is a product-name; Queue?
 
-A queue in product-name; is an ordered collection of messages. Messages are enqueued and dequeued (delivered to consumers) in a ([FIFO ("first in, first out")](https://en.wikipedia.org/wiki/FIFO_(computing_and_electronics)) manner. Refer to [what is a queue?](##whatisaqueue) for a more detailed definition of a queue in general. 
+A queue in product-name; is an ordered collection of messages. Messages are enqueued and dequeued (delivered to consumers) in a ([FIFO ("first in, first out")](https://en.wikipedia.org/wiki/FIFO_(computing_and_electronics)) manner. Refer to [what is a queue?](#whatisaqueue) for a more detailed definition of a queue in general. 
 
 Many features in a messaging system are related to queues. Some RabbitMQ queue features such as priorities and [requeueing](./confirms.html) by consumers can affect the ordering as observed by consumers. 
 
--- a/site/queues.md
+++ b/site/queues.md
@@ -15,11 +15,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-# product-name; Queues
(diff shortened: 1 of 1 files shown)


site/queues.md
```diff
@@ -15,59 +15,51 @@ See the License for the specific language governing permissions and
...
+# product-name; Queues
 
-## Introduction
+## What is a product-name; Queue?
 
-This guide provides an overview of queues in RabbitMQ. Since
-many features in a messaging system are related to queues, it
-is not meant to be an exhaustive guide but rather an overview
-that provides links to other guides.
+A queue in product-name; is an ordered collection of messages. Messages are enqueued and dequeued (delivered to consumers) in a ([FIFO ("first in, first out")](https://en.wikipedia.org/wiki/FIFO_(computing_and_electronics)) manner. Refer to [what is a queue?](##whatisaqueue) for a more detailed definition of a queue in general.
```
dyozie, line 22: Did we decide to try and limit usage of product-name; to the titles?

dyozie: Also:

##whatisaqueue -> #whatisaqueue

dyozie (approved):

pstack2021:
Hey @michaelklishin,

Can you approve this PR if you are happy with the updates. Also, as per my slack conversation: https://vmware.slack.com/archives/C03SXCHA7PD/p1692701370485379 I am not adding the product name to topic titles now so I think the remaining improvements I made to this topic should be ok - many thanks in advance.

michaelklishin (approved):

PR #1717, opened 2023-08-24
SEO updates in lazy queue content

pstack2021:
Hi @acogoluegnes,

Can you please review this PR. The main of this PR is to do the following: 

1. Provide a clear crisp title (returned in search engine result pages (SERPs)
2. Define a classic queue in "lazy" queue mode (otherwise known as a lazy queue) - do this is the first few sentences (this is what readers will see in the SERPs)
3. Make crystal clear the recommendation is to use quorum queues instead 
4. Next is a "Deeper Dive" into lazy queues

I didn't change any other information in this topic.

Question on this content: 

_By default, queues keep an in-memory cache of messages that is filled up as messages are published into RabbitMQ. The idea of this cache is to be able to deliver messages to consumers as fast as possible. Note that persistent messages can be written to disk as they enter the broker and kept in RAM at the same time._

Q. 1. _By default, queues_  I presume you are just referring to classic queues here?
Q.2  _queues keep an in-memory cache of messages......._ is this non-persistent messages? if so should we specify this as we refer to persistent messages in the next sentence. 

Many thanks in advance

### DIFF
--- a/site/lazy-queues.md
+++ b/site/lazy-queues.md
@@ -1,40 +1,43 @@
-# Lazy Queues
+# Classic Queues Operating in "Lazy" Queue Mode (A Lazy Queue)
 
-## <a id="interstitial" class="anchor" href="#interstitial">Wait, There's a Better Way: Next Generation Classic Queue Storage</a>
+## What is a Lazy Queue
 
-This guide covers lazy mode of classic queues.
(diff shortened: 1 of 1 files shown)


michaelklishin:
These are good writing improvements in general.

Thank you, @pstack2021!

PR #1728, opened 2023-09-21
Classic Queue version at queue declare

SignalWhisperer:
The RabbitMQ website does not have any mention of being able to set the queue version for Classic Queues at queue declaration. This change adds a short description on how to do so through `x-queue-version`, which is only mentioned in the following [RabbitMQ 3.12 Performance Improvements blog post](https://blog.rabbitmq.com/posts/2023/05/rabbitmq-3.12-performance-improvements/). This way of controlling the feature was however available since the launch of CQv2 in 3.10.0 as far as I can tell.

The phrasing used is lifted from the [lazy queues documentation](https://www.rabbitmq.com/lazy-queues.html#configuration) to try and maintain consistency across the website.

### DIFF
--- a/site/classic-queues.md
+++ b/site/classic-queues.md
@@ -171,6 +171,10 @@ The default version can be set through configuration by setting
 only affects newly declared queues. Pre-existing queues will remain on version 1,
 until explicitly migrated or deleted and redeclared.
 
+The queue version can also be set by supplying the `x-queue-version`
+queue declaration argument with an integer specifying the desired
+version.
+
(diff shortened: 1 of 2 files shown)


michaelklishin:
Thank you for taking a moment to contribute to RabbitMQ.

This is by design. The docs do mention that you can [use a policy](https://rabbitmq.com/parameters.html#why-policies-exist) and you should always consider that option first over optional queue arguments.

Your future self will thank you for making this choice.

PR #1734, opened 2023-09-28
Update which-erlang.md

johanrhodin:
This update make the Erlang versions match what is in the release notes for RabbitMQ `3.12.0`->`3.12.6`.


michaelklishin:
This breaks page rendering:

```
lxml.etree.XMLSyntaxError: Opening and ending tag mismatch: tr line 96 and table, line 391, column 9
```

PR #1735, opened 2023-09-29
Add stream replication port info

dreiucker:
We consulted the documentation, but only found later during testing that for stream replication requires specific stream replication ports. Since AMQP ports are used for the core stream it was unexpected that another port range was added for stream replication. Next time we will check the network guide for sure.

May not concern all developers, but usually network ports have to be explicitly enabled. So hoping that entry makes the requirement more obvious for people that use and support the product.


michaelklishin:
Thank you!

PR #1739, opened 2023-10-10
Clarify feature flags documentation

dumbbell:
Feature flags are never removed. Thus, this patch reworks sections which talk about removed feature flags to instead explain required feature flags.

Also, use the notion of "stable feature flags" instead of "supported feature flags". They are not the same. Experimental feature flags are supported too, but they are not enabled by default.

Finally, rework the section about `$RABBITMQ_FEATURE_FLAGS` to explain when it is useful and that required feature flags are still enabled.

### DIFF
--- a/site/feature-flags.md
+++ b/site/feature-flags.md
@@ -122,20 +122,23 @@ for the following reasons:
 
  * Skipping minor versions is not tested in CI.
  * Non-sequential releases may or may not support the same set of feature
-   flags. Support for older feature flags can be removed. Flag present for
-   several minor branches, they are removed and their associated
-   feature/behavior is now implicitly enabled by default, preventing clustering
-   with older nodes. Feature flags are kept around for a number (say, two) of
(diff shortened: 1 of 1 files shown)


site/feature-flags.md
```diff
@@ -17,31 +17,34 @@ to determine if they are compatible and then communicate together,
 despite having different versions and thus potentially having different
 feature sets or implementation details.
 
-This subsystem was introduced in to allow for **[rolling upgrades](./upgrade.html#rolling-upgrades)
-of cluster members without shutting down the entire cluster**.
+This subsystem was introduced in to allow for **[rolling
```
mkuratczyk, line 20:
```suggestion
This subsystem was introduced to allow for **[rolling
```

mkuratczyk (approved):

site/feature-flags.md
dumbbell: Thank you, fixed!

PR #1740, opened 2023-10-11
change redirect to stderr because golang log outputs stderr.

c3-suzuki:
I followed the tutorial on the page below, but the logs were not written to the file. In the case of Golang, `log.Printf()` outputs to stderr, so I suggest changing the redirect to stderr.

https://www.rabbitmq.com/tutorials/tutorial-four-go.html

>
> If you want to save only 'warning' and 'error' (and not 'info') log messages to a file, just open a console and type:
> ~~~
> go run receive_logs_direct.go warning error > logs_from_rabbit.log
> ~~~


michaelklishin:
Good catch, I could reproduce this behavior. Thank you!

PR #1743, opened 2023-10-17
fix: wrong argument types and recommended consume method in php tutorial

ramunasd:
A few small fixes in PHP tutorial:

* fix wrong argument type `basic_qos`
* recommended method for consuming messages
* object orientied way to get message properties
* require more recent library version


michaelklishin:
@ramunasd thank you. Can you please see if any of the [executable tutorials](https://github.com/rabbitmq/rabbitmq-tutorials/) may need updating, too?

michaelklishin (approved):

PR #1745, opened 2023-10-22
Path is required for ProxyPass directive in Apache configuration

timcu:
Path is required for ProxyPass directive and ProxyPassReverse directive in Apache configuration unless inside a <Location> section which rabbitmq management example is not.

https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypass

Further down the rabbitmq management page, in a different section, the correct syntax is used https://www.rabbitmq.com/management.html#proxy

Also added that both mod_proxy and mod_proxy_http are required to be enabled for this to work.

### DIFF
--- a/site/management.md
+++ b/site/management.md
@@ -494,12 +494,12 @@ To support encoded slashes in URIs, Apache requires users to explicitly enable
 AllowEncodedSlashes On
 </pre>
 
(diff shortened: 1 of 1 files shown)


michaelklishin (approved):

michaelklishin:
Thank you!

PR #1751, opened 2023-11-06
Update wording and arrangement in `streams.md`

adamball:
I was reading the Streams documentation and thought the changes in this PR would aid in ease of reading.

The content itself has only been slightly adjusted.


michaelklishin:
Thank you!

PR #1764, opened 2023-11-17
Update tutorial-three-go.md

jvansan:
In Go, `log.Printf` uses STDERR, so need to capture both STDOUT + STDERR for the logging to actually show up in a file.

### DIFF
--- a/site/tutorials/tutorial-three-go.md
+++ b/site/tutorials/tutorial-three-go.md
@@ -471,7 +471,7 @@ func main() {
 If you want to save logs to a file, just open a console and type:
 
 <pre class="lang-bash">
-go run receive_logs.go > logs_from_rabbit.log
+go run receive_logs.go &> logs_from_rabbit.log
 </pre>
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thanks. That likely will need special escaping, let me see.

jvansan:
I'm not sure if this matters, but in the next [Go tutorial](https://www.rabbitmq.com/tutorials/tutorial-four-go.html), you specicially use `go run receive_logs_direct.go warning error 2> logs_from_rabbit.log`, ie capture STDERR only, where as the `&` captures both. It might be clearer to be consistent.

michaelklishin:
@jvansan capturing both is what I'd do. You are welcome to submit a PR but keep in mind that `&` must be escaped for DocBook-safe rendering (like in https://github.com/rabbitmq/rabbitmq-website/commit/6408eb6619b479306eff4e7e9c6f7e0780691e25).

PR #1770, opened 2023-11-30
Hint for kernel page chache problem in older kubernetes versions

Kraego:
This note would have saved me several hours of research... PR is ready when update of kubernetes fixes the problem


michaelklishin:
@Kraego if you suggest that we mention that the K8S/cgroups combination that reports kernel page cache as "memory used by the pod" (app), which is not really correct, we can do it today.

Kubernetes [1.25 with cgroups v2 support](https://kubernetes.io/blog/2022/08/23/kubernetes-v1-25-release/#support-for-cgroups-v2-graduates-to-stable) has been out for more than a year.

Any reasons why we should not make the docs say something like "this affects Kubernetes installations older than 1.25 and Linux distributions that use cgroups v1"?

michaelklishin:
I've edited this a bit and manually merged it into `live`

PR #1778, opened 2023-12-18
Update default exchange restrictions

Zerpet:
RabbitMQ does not allow binding operations on the default exchange. Such
operations result in a 403 - Not Allowed error, at the time of this
writing.

We conform with the spec in the creation of the default exchange, and
default bindings i.e. every queue is implicitly bound to the default
exchange with routing key equal to queue name.

### DIFF
--- a/site/tutorials/amqp-concepts.md
+++ b/site/tutorials/amqp-concepts.md
@@ -160,6 +160,9 @@ other words, the default exchange makes it seem like it
 is possible to deliver messages directly to queues, even
 though that is not technically what is happening.
 
+The default exchange, in RabbitMQ, does not allow bind/unbind operations.
+Binding operations to the default exchange will result in an error.
+
 ### <a id="exchange-direct" class="anchor" href="#exchange-direct">Direct Exchange</a>
 
 A direct exchange delivers messages to queues based on the
--- a/site/specification.xml
+++ b/site/specification.xml
@@ -1068,7 +1068,7 @@ limitations under the License.
           </rule>
 
           <rule>
-            <status value="ok"/>
+            <status value="doesn't"/>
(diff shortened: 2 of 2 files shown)


Zerpet:
Related to https://github.com/rabbitmq/amqp091-go/issues/231

site/tutorials/amqp-concepts.md
```diff
@@ -160,6 +160,9 @@ other words, the default exchange makes it seem like it
 is possible to deliver messages directly to queues, even
 though that is not technically what is happening.
 
+The default exchange, in RabbitMQ, does not allow bind/unbind operations.
+Binding operations to the default exchange will result in an error.
```
mkuratczyk, line 164:
```suggestion
Binding operations to the default exchange will result in an error.

In RabbitMQ, the default exchange is a specialised direct exchange which uses the routing key as the queue name.
```

mkuratczyk (changes requested):
I'd add a sentence to explain what the default exchange is. otherwise it sounds like it serves no purpose (what do you need an exchange for if it can't having bindings)

site/tutorials/amqp-concepts.md
Zerpet: That's explained in the first paragraph of this section (it doesn't show in the git diff):

> The default exchange is a direct exchange with no name
(empty string) pre-declared by the broker. It has one special
property that makes it very useful for simple applications:
every queue that is created is automatically bound to it with
a routing key which is the same as the queue name.

Do you think we should re-iterate the same here?

mkuratczyk: oh, no need then

mkuratczyk (approved):

PR #1780, opened 2023-12-20
Fix RABBITMQ_MNESIA_DIR description in relocate docs

paultcochrane:
While reading through the relocate docs, I noticed that a word seemed to be missing from the `RABBITMQ_MNESIA_DIR` description.  A quick `git grep` showed what the missing word should be from a similar description in the configure docs.  This change makes the description for this environment variable internally consistent.

This PR is submitted in the hope that it is useful. If you wish for any changes, please let me know and I'll be happy to update as necessary and resubmit.


michaelklishin:
Solid investigative skills. Thank you!

PR #1784, opened 2024-01-12
Fix advanced.config example on quorum queue page

rolatsch:
Add missing point to advanced.config example so it can be copied and pasted.


michaelklishin (approved):

michaelklishin:
Well spotted, thank you!

PR #1792, opened 2024-01-24
Replace Message with Messaging

schelv:
Replace Message with Messaging.
That is what is used everywhere else in the document.


schelv:

PR #1793, opened 2024-01-28
Fix Runtastic link

firefly-cpp:
**Runtastic** case study on the Medium platform is no longer available. Thus, I added a reference to the home page of **Runtastic** for now.


michaelklishin:
The goal was to link to a specific blog post from the Runtastic engineering team. If that's now gone, linking to the app won't really solve anything.

I could not find a new home for their blog, so perhaps post-acquisition something has changed around their engineering team.

michaelklishin:
I guess

PR #1799, opened 2024-02-19
Update production-checklist.md: bit/byte units

dchaley:
Hi RabbitMQ 👋🏻 

I was going through the [production checklist](https://www.rabbitmq.com/production-checklist.html)–this is awesome!–and I think the network throughput units are slightly wrong. I think this corrects & clarifies the units.

Changes:
- Use KB (1000 bytes) instead of kiB (1024 bytes). Example source: https://en.wikipedia.org/wiki/Kilobyte


dchaley:

michaelklishin:
Thank you!

PR #1803, opened 2024-02-22
replace &lt; with <

bigzooooz:
since direct copy/paste would result in `Syntax error: "&" unexpected`, I suggest updating the codes in docs


michaelklishin:
In the previous version of the site (that is, what we had between 2007 and literally yesterday), the escape sequences were necessary.

Thanks, it's a good catch.

PR #1806, opened 2024-02-25
doc(tutorials): improve go instructions

vit0rr:
The goal is to improve the instructions by adding how to init a mod with Go, creating the `main` function, and adding a note to create a new directory for the receiver. This is necessary because, in GoLang, we cannot have two files with the same function name in the same directory. We cannot create two files with the same `func main(){}` name in this case.

It can improve the tutorial's clarity and prevent the reader from having problems with the module and the package. Because you need to be inside a module to run `go get`.

### DIFF
--- a/tutorials/tutorial-one-go.md
+++ b/tutorials/tutorial-one-go.md
@@ -57,11 +57,11 @@ on behalf of the consumer.
 > First, install amqp using `go get`:
 >
 > ```go
+> go mod init <your-module-name>
 > go get github.com/rabbitmq/amqp091-go
 > ```
 
-Now we have amqp installed, we can write some
-code.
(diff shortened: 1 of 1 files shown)


tutorials/tutorial-one-go.md
```diff
@@ -192,6 +194,8 @@ func failOnError(err error, msg string) {
 }
 ```
 
+Create a new directory for the receiver, like `receiver/receive.go, to avoid [DuplicateDecl](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#DuplicateDecl) error.
```
michaelklishin, line 197: The term "consumer" is used in this and other tutorials.

PR #1820, opened 2024-03-05
Changing the Introduction

pstack2021:
Why:
The [current introduction page ](https://www.rabbitmq.com/docs/documentation) is a table of contents, now that we have a doc table of contents on the new site, we don't need to list links to all the docs on this page. Updating now to provide users with a docs welcome page to help them navigate around this new versioned documentation site.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -2,7 +2,7 @@
 title: RabbitMQ Documentation
 ---
 <!--
-Copyright (c) 2005-2024 Broadcom. All Rights Reserved. The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
+Copyright (c) 2007-2024 Broadcom. All Rights Reserved. The term "Broadcom" refers to Broadcom Inc. and/or its subsidiaries.
 
 All rights reserved. This program and the accompanying materials
 are made available under the terms of the under the Apache License,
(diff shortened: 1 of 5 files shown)


pstack2021:
@dumbbell 

Note, the links to release information and getting started pages are links to this site currently: https://revamp-the-rabbitmq-home-pag.rabbitmq-website.pages.dev/ I know these need to change.

Also what are your thoughts on including a table on this page for the different personas i.e. Developer, Administration ....1 column per persona, then under each persona we could include a linked list to the documentation that each persona would use setup RabbitMQ for that persona??? 

Thanks

dumbbell:
Starting to organize topics based on the reader’s profile is great! I believe this should be the primary focus of this page.

Content such as:
* RabbitMQ Documentation is now versioned
* which is currently 3.13
* Select the 3.12 option from the drop-down menu to access documentation for previous 3.12.x releases
* Soon we will be revising the table of contents
* We will keep you posted!
… is ephemeral because it will be obsolete as soon as there is a new version or when we change the ToC. This phrasing is better suited for a blog post. Furthermore, because that page is versioned, maintaining ephemeral content will be difficult.

I think the welcome page should remain valid at any time. If we focus on administrators and developers, the page will greatly help people navigate the docs.

pstack2021:
Thanks JSP for your feedback, I will add some sections for user profiles. With regard to your comments above, I agree with them yes but I think for now on a new doc site, which users are navigating around/getting used to lets include this information. Yes, it will change in time (new release/changes in TOC) but that won't be for a while....the next RabbitMQ release is not until 4.0 and that is not due out for a long time yet.

pstack2021:
also documentation.md needs to be renamed to index.md before merging.

docs/documentation.md
```diff
@@ -17,6 +17,111 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+  <tr>
+    <td class="centre">Install and configure additional plugins</td>
+    <td class="centre">XXX</td>
+  </tr>
+  <tr>
+    <td class="centre">Choose the appropriate Queue type to use and configure the Queues</td>
+    <td class="centre">XXX</td>
+  </tr>
+  <tr>
+    <td class="centre">Configure publishers and consumers</td>
```
michaelklishin, line 95: Administrators do not configure publishers and consumers, applications do, at least applications do the most important parts. This belongs to the developer section.

docs/documentation.md
```diff
@@ -17,6 +17,111 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+  <tr>
+    <td class="centre">Install and provision RabbitMQ on all Nodes in the Cluster</td>
+    <td class="centre">XXX</td>
+  </tr>
+  <tr>
+    <td class="centre">Configure the Hostnames</td>
+    <td class="centre">XXX</td>
+  </tr>
+  <tr>
+    <td class="centre">Configure High Availability (if required)</td>
```
michaelklishin, line 75: This should not link to the classic mirrored queues guide. I'd rename it to "Replicated Queues and Streams."

docs/documentation.md
```diff
@@ -17,6 +17,111 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+   <tr>
+    <th class="centre">Tasks</th>
+    <th class="centre">Documentation Reference</th>
+  </tr>
+  <tr>
+    <td class="centre">Install and provision RabbitMQ on all Nodes in the Cluster</td>
+    <td class="centre">XXX</td>
+  </tr>
+  <tr>
+    <td class="centre">Configure the Hostnames</td>
```
michaelklishin, line 71: With many deployment options you don't have to configure hostnames, the only thing RabbitMQ needs is that the hostnames are resolvable on all nodes and stable.

This is a part of the broader networking configuration, and IIRC is documented in the Networking and Clustering guides.

"Forming a cluster" might be a better option with Clustering, Cluster Formation and Networking guides being the three key guides.

docs/documentation.md
```diff
@@ -17,6 +17,111 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+For an administrator who want to use RabbitMQ in a production environment, the following is the generic list of tasks they would follow to start using RabbitMQ.
+
+<table id="administrator">
+   <tr>
+    <th class="centre">Tasks</th>
+    <th class="centre">Documentation Reference</th>
+  </tr>
+  <tr>
+    <td class="centre">Install and provision RabbitMQ on all Nodes in the Cluster</td>
```
michaelklishin, line 67: I find the "on all nodes in the cluster" part to be redundant and confusing.

I'd say we have

1. Installation options
2. Kubernetes guides (because this is a specific and thoroughly documented installation option)
3. Cluster formation

docs/documentation.md
```diff
@@ -17,6 +17,111 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+Use the navigation on the left to browse through documentation for your release of RabbitMQ.
+
+## Getting Started
+Choose which RabbitMQ version you want to use, review the [latest RabbitMQ releases](https://revamp-the-rabbitmq-home-pag.rabbitmq-website.pages.dev/docs/changelog) to help you decide. Next, go to the [Get Started pages](https://revamp-the-rabbitmq-home-pag.rabbitmq-website.pages.dev/tutorials) to try out the tutorials. These tutorials offer an introduction to messaging and it's key features as well as some common usage scenarios.
+
+## User: Developer
+
+For a developer who wants to use RabbitMQ in an application, the following is the generic list of steps they would follow to start using RabbitMQ.
+
+<table id="developer">
```
michaelklishin, line 35: Tutorials, Connections, Channels, Publishers, Consumers, TLS configuration, Reliability, Clustering, Quorum Queues, Streams, Heartbeats, TTL, Dead Lettering, Priority Queues, MQTT, STOMP, and possibly the three Troubleshooting are what I'd consider "Developer guides."

docs/documentation.md
```diff
@@ -17,6 +17,111 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+# RabbitMQ Documentation
+
+Welcome to RabbitMQ documentation. Use the drop-down menu on the top right of any page to select the documentation version for the RabbitMQ release that you are using. We update each version of documentation with the latest minor and patch information for that release.
+
+For previous 3.12.x documentation, select the **3.12** option from the drop-down menu. 3.12.x documentation is hosted on our previous RabbitMQ site. This documentation is not versioned. 
+
+Use the navigation on the left to browse through documentation for your release of RabbitMQ.
+
+## Getting Started
+Choose which RabbitMQ version you want to use, review the [latest RabbitMQ releases](https://revamp-the-rabbitmq-home-pag.rabbitmq-website.pages.dev/docs/changelog) to help you decide. Next, go to the [Get Started pages](https://revamp-the-rabbitmq-home-pag.rabbitmq-website.pages.dev/tutorials) to try out the tutorials. These tutorials offer an introduction to messaging and it's key features as well as some common usage scenarios.
```
michaelklishin, line 29: We must use `./relative/paths/` here, we cannot be linking to `rabbitmq-website.pages.dev` as some of these will certainly be overlooked and deployed live. We've had it with website indexing, for example.

pstack2021: Sure @michaelklishin will do, I am just using these links temporarily, thanks.

pstack2021:
Hi @dumbbell and @mkuratczyk 

Can you please review the new doc intro page. Please note this is the initial DRAFT, I need your input now with the list of tasks each persona (admin and developer) would complete to use RabbitMQ. I know it is difficult to do this as the route each persona takes is very much dependant on their own requirements. In the important notes before the tables I am attempting to reiterate this.

The layout i.e. current tabular format in HTML will all change as will the the link format (@dumbbell will help me with this when the time comes) so please don't comment on that for now. Right now, we must get the content correct i.e. the general list of tasks a developer and an administrator would complete

Please feel free to update in the branch itself. Also, this `documentation.md `file needs to be renamed to` index.md `before it goes live.

Another point @dumbbell, when we have the information for these 2 personas in decent shape, perhaps we can link/reference your use cases: https://www.rabbitmq.com/ from this intro page i.e. match up a persona with one of those particular use cases? 

Many thanks in advance,
Paula.

docs/documentation.md
```diff
@@ -17,6 +17,124 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
+# RabbitMQ Documentation
+
+Welcome to RabbitMQ documentation. Use the drop-down menu on the top right of any page to select the documentation version for the RabbitMQ release that you are using. We update each version of documentation with the latest minor and patch information for that release.
+
+For previous 3.12.x documentation, select the **3.12** option from the drop-down menu. 3.12.x documentation is hosted on our previous RabbitMQ site. This documentation is not versioned.
```
dumbbell, line 24: I don't think we need this paragraph as the previous one already covered the fact that other versions are available in the list.

Also, "This documentation is not versioned" is not true. It is versioned: the website contains the docs for 3.12.x (but only that one).

docs/documentation.md
```diff
@@ -17,6 +17,124 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+    <td class="centre">Start by going through the tutorials</td>
+    <td class="centre"><a href="https://www.rabbitmq.com/tutorials">Getting Started</a></td>
+  </tr>
+  <tr>
+    <td class="centre">Choose a messaging protocol</td>
+    <td class="centre"><a href="https://www.rabbitmq.com/protocols.html">What Protocols does RabbitMQ support</a></td>
+  </tr>
+  <tr>
+    <td class="centre">Open a Connection</td>
+    <td class="centre"><a href="https://www.rabbitmq.com/connections.html">Connections</a></td>
```
dumbbell, line 47: I feel this is a client library matter, not a server one at this point. I mean, the step is relevant, but perhaps it should point to the first tutorial.

To be honest, the whole "Getting started" section will have to be revisited to take into account what RabbitMQ is today.

docs/documentation.md
```diff
@@ -17,6 +17,124 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+  </tr>
+   <tr>
+   <td class="centre">Publish and Consume Messages </td>
+    <td class="centre">
+    <ul>
+      <li><a href="./publishers.md">Publishing Messages</a></li>
+      <li><a href="./publishers.md">Consuming Message</a></li>
+      <li><a href="./confirms.md">Consumer Acknowledgements and Publisher Confirms</a></li>
+    </ul>
+   </td>
```
dumbbell, line 57: After looking at this again, the list is very close to what is tought by the tutorials (even though they are AMQP 0-9-1 centric).

Thus I'm wondering if the first step should point to tutorials, then move to intermediate/advanced topics directly.

docs/documentation.md
```diff
@@ -17,6 +17,124 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+    <th class="centre">Documentation</th>
+  </tr>
+   <tr>
+    <td class="centre">Before deploying RabbitMQ in a production environment, take a look at our deployment guidelines</td>
+    <td class="centre"><a href="https://www.rabbitmq.com/docs/download">Deployment Guidelines</a></td>
+  </tr>
+  <tr>
+  <tr>
+    <td class="centre">Choose an installation option for your platform. When you are deploying RabbitMQ, choose one of the latest RabbitMQ and Erlang versions.</td>
+    <td class="centre"><a href="https://www.rabbitmq.com/docs/download">Installing RabbitMQ</a> and <a href="https://www.rabbitmq.com/docs/platforms">Supported Platforms</a></td>
```
dumbbell, line 108: "Supported Platforms" should be linked from the "Installing RabbitMQ" page. The "supported" notion is implicit in "installing": the user would click to learn how to install and will look for their platform in that page.

dumbbell (commented):
My general feeling after taking a step back is that the tables generally repeat the table of content. You did a great job with that ToC and it's well organized and already reflects the questions a user (developer or administrator) will have and in the correct order. I'm having second thoughts about a list of tasks each user may go through.

I mean:
* We already have tutorials for developers that cover the basics. After the tutorials, it's hard to predict what the developer will look for because it will be dependant on their use case.
* For admins, the sequence is usually install -> configure -> backup/monitor. Underneath, the order is dependent on the use case as well.

I'm wondering if we should keep the developer and admin sections, but have a simpler paragraph guiding each persona to find the entry point (tutorials for developers, the install page for admins).

pstack2021:
> My general feeling after taking a step back is that the tables generally repeat the table of content. You did a great job with that ToC and it's well organized and already reflects the questions a user (developer or administrator) will have and in the correct order. I'm having second thoughts about a list of tasks each user may go through.
> 
> I mean:
> 
> * We already have tutorials for developers that cover the basics. After the tutorials, it's hard to predict what the developer will look for because it will be dependant on their use case.
> * For admins, the sequence is usually install -> configure -> backup/monitor. Underneath, the order is dependent on the use case as well.
> 
> I'm wondering if we should keep the developer and admin sections, but have a simpler paragraph guiding each persona to find the entry point (tutorials for developers, the install page for admins).

Thanks a lot @dumbbell for this feedback, I will take make additional changes to this page based on your comments. I shall have the next iteration out for review by the end of today, thanks.

pstack2021:
Hey @dumbbell

Can you take a look now, I removed the "Developer" and "Administrator" tables with the list of tasks and instead have 2 paragraphs about these personas under each persona heading.

Many thanks
Paula

docs/documentation.md
```diff
@@ -17,6 +17,24 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
+# RabbitMQ Documentation
+
+Welcome to RabbitMQ documentation. RabbitMQ is a reliable and mature messaging and streaming broker, which is easy to deploy on cloud environments, on-premises, and on your local machine. With support for many different messaging protocols, it can scale from providing point-to-point messaging all the way up to globally distributed highly-available event streaming with rock-solid reliability. RabbitMQ is currently used by millions worldwide. 
+
+## How to use this documentation
+
+Use the navigation on the left to browse through documentation for your release of RabbitMQ. This table of contents is structured so it can be used by our two main RabbitMQ personas: Developer and Administrator.
```
dumbbell, line 26: I think I preferred when the introduction was more to the point. That first paragraph is for the homepage. I mean, either the user is browsing the website to learn more about the project and probably started with the homapage, or they are here for the docs and would probably prefer to get the information they want straight away, not a general paragraph about something they are already using.

docs/documentation.md
```diff
@@ -17,6 +17,24 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+Welcome to RabbitMQ documentation. RabbitMQ is a reliable and mature messaging and streaming broker, which is easy to deploy on cloud environments, on-premises, and on your local machine. With support for many different messaging protocols, it can scale from providing point-to-point messaging all the way up to globally distributed highly-available event streaming with rock-solid reliability. RabbitMQ is currently used by millions worldwide. 
+
+## How to use this documentation
+
+Use the navigation on the left to browse through documentation for your release of RabbitMQ. This table of contents is structured so it can be used by our two main RabbitMQ personas: Developer and Administrator. 
+
+## User: Developer
+
+Each developer scenario is different depending on the use case. For example, lets take the task of opening a connection, this route differs a lot depending on the messaging protocol a developers uses. This is just one factor, there are many more. You might want to start with [Getting Started](/tutorials), which is a set of tutorials that will guide you on how to use RabbitMQ with AMQP. If you need more details, then go the **How to Use RabbitMQ** section.
```
dumbbell, line 30: I would skip the first part which is kind of obvious to the reader IMHO.

The paragraph would start at the "You might want to (...)", pointing the reader to the tutorials if they are new to RabbitMQ. Then talk about the "How to use RabbitMQ" section as you do in the last sentence.

docs/documentation.md
```diff
@@ -17,6 +17,24 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+## User: Administrator
+
+Each administrator scenario is also different depending on the use case. For administrators, the sequence might usually be install, configure, and monitor. As a recommendation only, start with **Install and Upgrade** documentation on the left, then access the **How to Manage RabbitMQ** tasks, and finally **How to Monitor RabbitMQ**.
```
dumbbell, line 34: Likewise, I would go directly to the reasoning behind the ToC:
"The docs is divided into two sections:
* How to manage RabbitMQ where you can learn about the configuration of the broker
* How to monitor RabbitMQ where we describe best practices to keep an eye on the service health"
(this phrasing can probably be improved :)

docs/documentation.md
```diff
@@ -17,6 +17,24 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 -->
+# RabbitMQ Documentation
+
+Welcome to RabbitMQ documentation. RabbitMQ is a reliable and mature messaging and streaming broker, which is easy to deploy on cloud environments, on-premises, and on your local machine. With support for many different messaging protocols, it can scale from providing point-to-point messaging all the way up to globally distributed highly-available event streaming with rock-solid reliability. RabbitMQ is currently used by millions worldwide. 
+
+## How to use this documentation
+
+Use the navigation on the left to browse through documentation for your release of RabbitMQ. This table of contents is structured so it can be used by our two main RabbitMQ personas: Developer and Administrator. 
+
+## User: Developer
```
dumbbell, line 28: What about a more "inviting" title like "If you are a developer"?

dumbbell:
@pstack2021: I applied the latest changes we discussed and rebased the branch on top of the latest main.

In addition, I used the CSS style sheet to change the paragraphs at the beginning that talks about the drop-down menu and the table of contents. On large screens, they point to the top-right corner and the left sidebar and have icons to make this even more visible. On small screen, when the left sidebar disappears, the paragraphs are replaced to mention the hamburger icon and the main menu. So now, they adapt to what is actually displayed for the user.

The preview is available at the same URL:
https://revamping-docs-introduction.rabbitmq-website.pages.dev/docs

pstack2021:
Thanks a lot @dumbbell

PR #1821, opened 2024-03-05
docs: quorum queues and dlx: fix anchors wrt 'poison message handling' details

geor-g:
See the commit messages for context and details.


michaelklishin:
Thank you. I think we should customize the anchor because "for-quorum-queues" at the end sounds overly repetitive to me.

I'll get it now and credit you for the finding.

PR #1822, opened 2024-03-06
Fixed typo in docs/Prometheus

ArjunMG:


michaelklishin:
I'm afraid we don't accept automated PRs like this. There are no reasons to change list bullets in that or any other guide from `*` to `-`, for example.

michaelklishin:
To be clear, we would accept and we do appreciate typo fixes. But this PR changes way too much, including things that are not typos by any stretch.

PR #1825, opened 2024-03-07
Avoid escaping XML inside xml code block

jipe:
The XML to be used for pom.xml defined in an XML code block in markdown was made using XML entities for brackets which resulted in them also being escaped in the resulting view.


michaelklishin:
Thank you. Yes, these were necessary with the original site but not with Docosaurus.

PR #1807, opened 2024-02-26
Update tutorial-two-spring-amqp.md

eshonkulov-asliddin:
Fixed typo

PR #1827, opened 2024-03-07
Fixed Markdown formatting issue with html character entities

MoellJ:
## Trivial / obvious fix

Some tutorials contained example code that used the html character entities `&lt;` and `&gt;` instead of `< >`.
Similarly to #1807 this PR aims to fix some of these issues.

Other possible erronous files that I was not sure about and therefore did not edit:
- tutorial-five-javascript.md
- Many php tutorials
- Many objective-C tutorials
- Other Markdown files containing &lt; and &gt;


michaelklishin (approved):
Thank you! These were necessary with the original website but not with Docosaurus 👍

PR #1833, opened 2024-03-11
Mention delete_queue params ignored by Streams

ikavgo:
see #10543


michaelklishin:
This file is generated from [its `man` equivalent](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/docs/rabbitmqctl.8). I can add take care of that if you don't want to figure out its really obscure format, troff/ntroff.

ikavgo:
I can do this, thank you

PR #1843, opened 2024-03-16
More improvements to release information

dumbbell:
This is a collection of small improvements:
* The table’s column titles are ajusted, following feedback from @pstack2021.
* The table’s legend is reworked to display correctly on small screens.
* The release data in `docusaurus.config.js` is reorganized to be able to record and show all patch releases in a release branch.


pstack2021:
Hey @dumbbell, 
Looks great thank you, 3 comments:
1. Please see this screenshot, I am viewing this page on the 14 inch macbook pro screen, as you can see, I need to scroll if I want to see the support columns which I guess isn't an ideal experience on a laptop, it seems to be the "Latest Patch" column that is cause the issue: "Release Notes" and "Announcement" for 3.13.0 are side by side, could we have a unordered list here with "Release Notes" first (on top)
 and "Annoucement" underneath it ...or we don't need an unordered list just place one under the order making this column slimmer and therefore allowing a user to see the full table? 

Screenshot 2024-03-18 at 21 58 54

2. Spelling Typo: [Annoucement](https://787ae2ae.rabbitmq-website.pages.dev/blog/2024/03/11/rabbitmq-3.13.0-announcement) should be "Announcement"

3. See screenshot below could have these are 2 seperate columns with the full column names i.e. "End of Community Support" and "End of Extended Commercial Support" like it was in the past, I think its clearer than "End of Support" and then "Community" and "Ext Commercial" underneath it, I totally understand why you did it to reduce column size so leave as is if it going to cause layout issues, if it is remaining the same, maybe we can change "Ext Commercial" to its full description which is "Extended Commercial" 

Screenshot 2024-03-18 at 22 56 09

pstack2021:
> @pstack2021: Here is how it would look like with full column names: image
> 
> What do you think?

Yes I like to @dumbbell, its 100% clear

PR #1860, opened 2024-03-29
Remove duplicated sidebar link (Stream Plugin)

naramdash:
- Remove duplicated `Stream Plugin` link.


dumbbell (approved):
Good catch! Thank you!

PR #1865, opened 2024-04-02
Federation: queue-type and resource-cleanup-mode

mkuratczyk:
* queue-type is a new field added in 3.13.1
* resource-cleanup-mode has been there for a long time but was not documented

### DIFF
--- a/docs/federation-reference.md
+++ b/docs/federation-reference.md
@@ -195,6 +195,35 @@ The following upstream parameters are only applicable to <a href="./federated-ex
     </tr>
 
     <tr>
+      <td><code>queue-type</code></td>
+      <td>
+        The queue type of the <a href="./federated-exchanges#implementation">upstream queue</a>. 
+
+        Defaults to `classic`. Set to `quorum` for high availability.
+
+        Changing the queue type will delete and recreate the upstream queue by default.
+        This may lead to messages getting lost (if they were in the upstream queue)
+        or missed (when they are published while there's no upstream queue to collect them).
+        You can set <code>resource-cleanup-mode</code> to <code>never</code> to avoid that.
+        This requires manually deleting the old upstream queue so that it can be recreated with
+        the new type.
+
+        Available since: 3.13.1
+      </td>
+    </tr>
+
+    <tr>
+      <td><code>resource-cleanup-mode</code></td>
+      <td>
+        Whether to delete the <a href="./federated-exchanges#implementation">upstream queue</a> when federation stops.
+
+        By default, the upstream queue is deleted immediately when federation link stops.
+        Set to `never` to keep the upstream queue around and collect messages even when
+        changing federation configuration.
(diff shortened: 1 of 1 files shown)


docs/federation-reference.md
```diff
@@ -194,6 +194,35 @@ The following upstream parameters are only applicable to <a href="./federated-ex
...
+      <td><code>queue-type</code></td>
+      <td>
+        The queue type of the <a href="./federated-exchanges#implementation">upstream queue</a>. 
+
+        Defaults to `classic`. Set to `quorum` for high availability.
+
+        Changing the queue type will delete and recreate the upstream queue by default.
+        This may lead to messages getting lost (if they were in the upstream queue)
+        or missed (when they are published while there's no upstream queue to collect them).
+        You can set <code>resource-cleanup-mode</code> to <code>never</code> to avoid that.
```
dumbbell, line 207: Could you please use Markdown when possible, instead of HTML (`<code>`)?

docs/federation-reference.md
```diff
@@ -194,6 +194,35 @@ The following upstream parameters are only applicable to <a href="./federated-ex
       </td>
     </tr>
 
+    <tr>
+      <td><code>queue-type</code></td>
+      <td>
+        The queue type of the <a href="./federated-exchanges#implementation">upstream queue</a>.
```
dumbbell, line 200: Could you please use Markdown links here.

docs/federation-reference.md
```diff
@@ -194,6 +194,35 @@ The following upstream parameters are only applicable to <a href="./federated-ex
...
+        You can set <code>resource-cleanup-mode</code> to <code>never</code> to avoid that.
+        This requires manually deleting the old upstream queue so that it can be recreated with
+        the new type.
+
+        Available since: 3.13.1
+      </td>
+    </tr>
+
+    <tr>
+      <td><code>resource-cleanup-mode</code></td>
```
dumbbell, line 216: Likewise.

docs/federation-reference.md
```diff
@@ -194,6 +194,35 @@ The following upstream parameters are only applicable to <a href="./federated-ex
...
+      <td>
+        Whether to delete the <a href="./federated-exchanges#implementation">upstream queue</a> when federation stops.
```
dumbbell, line 218: Likewise.

dumbbell (changes requested):

docs/federation-reference.md
michaelklishin: Curiously, this indeed works with Docusaurus but did not work with the original homegrown framework. We can then replace all the links in this guide with Markdown ones.

michaelklishin: By "this" I mean using Markdown links in an HTML block. Obviously Markdown links work just fine in a Markdown block.

PR #1710, opened 2023-08-16
Allow Oauth2 and basic auth for management UI and support multiple oauth2 resources

MarcialRosales:
**Note**: This PR should not be merged until the PR https://github.com/rabbitmq/rabbitmq-server/pull/10012 is released

PR #1866, opened 2024-04-02
Remove deprecated queue_explicit_gc_run_operation_threshold config notes

Ayanda-D:
Follow up to: https://github.com/rabbitmq/rabbitmq-server/pull/10880 
The deprecated config notes are misleading for users on 3.12.x or later.

### DIFF
--- a/docs/configure.md
+++ b/docs/configure.md
@@ -1482,45 +1482,6 @@ under the `rabbit` section.
       </p>
     </td>
   </tr>
-
-  <tr>
-    <td>
-      `lazy_queue_explicit_gc_run_operation_threshold`
-    </td>
-    <td>
-     Tunable value only for lazy queues when under memory pressure.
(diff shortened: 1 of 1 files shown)


michaelklishin:
This can only go into 4.x for which we currently do not have a documentation branch.

michaelklishin:
The earliest we can merge this is after https://github.com/rabbitmq/rabbitmq-website/pull/1710 is finished and the 3.13.x and 4.x branches are separated.

Ayanda-D:
FYI, a user wanted to apply/change this config in 3.12.7, then I double checked code only to find it removed/deprecated in 3.12.0. Hence the PRs. I'm not sure purpose of still having this config documented if no longer in use? Especially for lenghty period until 4.0 😄 hopefully doesnt cause further confusion while in the docs

PR #1871, opened 2024-04-04
release-information: List all patch releases

dumbbell:
## Why

With the switch to the old news+changelog pages to the new release information one, we lost the list of patch releases with their release date.

This commit restores the list of patch releases.

## How

To not bloat the screen/page, only the latest patch for each release branch is displayed.

There is an icon to let the user unfold the list and show all patch releases for a given release branch.

**V2**: The initial patch was only about the new support policy and especially the change of end of community support date change. This new patch keeps that but also lists all patch release for each release branch.

**V3**: Support small screens, with CSS tunings and new SVG icons.


src/components/RabbitMQServerReleaseInfo/index.js
```diff
@@ -89,32 +92,37 @@ export function RabbitMQServerReleaseInfoTable() {
...
+    var hasOSSSupport = false;
+    if (releaseBranch.end_of_support) {
+      /* Release branch is supported. */
+      if (previousReleaseBranch) {
+        const prevReleases = previousReleaseBranch.releases;
+        const initialPrevRelease = prevReleases[prevReleases.length - 1];
+        endOfSupportDates[0] = new Date(initialPrevRelease.release_date);
+      } else {
+        hasOSSSupport = true;
+        endOfSupportDates[0] = "Until next release branch";
```
michaelklishin, line 105: "Until next major or minor release" would be more specific

dumbbell: I tried exactly that first, but it's too long for the HTML table we use unfortunately (it breaks the readability of that table).

I plan to rework that table (and not use an HTML table in the first place) once #1710 is merged.

dumbbell: I changed it to "Until next minor/major". It's close to your suggestion while still being short to fit the table's constraints. What do you think?

michaelklishin (approved):

PR #1882, opened 2024-04-16
Add a "What's New" page for RabbitMQ 3.13

dumbbell:
The new page sums up what we already explained in details in several blog posts. It serves as a central place to find an overview of the new features of RabbitMQ 3.13.

The Release Information table is modified to link to this What's New page.


versioned_docs/version-3.13/whats-new.md
```diff
@@ -0,0 +1,81 @@
...
+
+MQTT is the standard protocol for the Internet of Things (IoT).
+
+IoT remote devices can have a poor network quality when connecting to the
+broker. Therefore, MQTT is lightweight: MQTT protocol headers are small to save
+network bandwidth. Since IoT devices might often disconnect and reconnect -
+imagine a car driving through a tunnel - MQTT is also efficient: Clients are
+connected and authenticated with a handshake that is shorter compared to other
+messaging protocols.
+
```
pstack2021, line 49: Could we restructure/rewrite this information slightly, I think the opening information under this heading should tell the user "Hey, RabbitMQ now supports MQTT 5.0, how great is that! I don't think we need to explain that MQTT is lightweight and the problems it solves users can pull that information from the net so how about: 

MQTT, which is the standard protocol for the Internet of Things (IoT) is supported by RabbitMQ for over a decade, and now our latest 3.13.0 release supports the latest version of MQTT, which is 5.0. The list of features that come with MQTT 5.0 to further address the issues of IoT device communication is vast. For a deeper dive into MQTT 5.0 support, its specifics with RabbitMQ, and the list of features you can avail of, go through the [MQTT Overview]( (/blog/2023/07/21/mqtt5) in our MQTT blog post.

versioned_docs/version-3.13/whats-new.md
```diff
@@ -0,0 +1,81 @@
...
+[MQTT 5.0](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html) is
+the latest version of the protocol and [is now supported by
+RabbitMQ 3.13.0](/blog/2023/07/21/mqtt5)!
+
+## Stream filtering
+
+[Stream filtering](/blog/2023/10/16/stream-filtering) allows to save bandwidth
+between the broker and consuming applications when those applications need only
+a subset of the messages of a stream.
+
```
pstack2021, line 59: I would get rid of the word "allows to" and just say "With the new Stream filtering feature, you can save bandwidth......"

versioned_docs/version-3.13/whats-new.md
```diff
@@ -0,0 +1,81 @@
...
+## Classic Queue Performance Improvements
+
+Classic queues can use [the newer message store
+v2](/blog/2024/01/11/3.13-release) which brings significant performance
+improvements, in terms of publishing/consuming rates, latency and memory
+footprint.
+
```
pstack2021, line 66: ok when I looked at this on preview: https://2667a207.rabbitmq-website.pages.dev/docs/whats-new I didn't know what this new message store was? the name of it etc, I had to go the blog to find out more, I then simply searched for "Message Store" .

Maybe under the title make the text that you have under this heading the 1st bullet point and then have a 2nd bullet point for "classic queue storage format version 1 is deprecated" and provide a one line how this results in classic queue performance improvements

versioned_docs/version-3.13/whats-new.md
```diff
@@ -0,0 +1,81 @@
...
+for RabbitMQ metadata that is designed to replace Mnesia. It is not yet *ready
+for production use* but we encourage users to try it out in test environments
+and provide feedback.
+
+Our plan is to completely remove Mnesia in the future. This should
+significantly improve RabbitMQ tolerance to network partitions. Once we switch
+to Khepri, there will be no partition handling strategy configuration
+(`pause_minority`, `autoheal`, and so on) — Khepri is based on the Raft
+protocol, just like quorum queues and therefore the semantics of what to do
+when a partition occurs are well defined and not configurable.
```
pstack2021, line 81: This section is very good. How about we add Khepri into the title " New Experimental Metadata Database: Khepri"

do we need "and therefore" in the last line........ how about just "Khepri is based on the Raft
protocol, just like quorum queues, the semantics of what to do
when a partition occurs are well defined and not configurable."

pstack2021: Are we documenting anything about "Message Containers" in the "What's new? I think this is more of an internal feature is it?

versioned_docs/version-3.13/whats-new.md
dumbbell: Indeed, that's better. I took your paragraph and modified the link at the end to say:
> go through [our MQTT 5.0 blog post]( (/blog/2023/07/21/mqtt5).

Because I think the entire blog post is useful, not just the overview section.

versioned_docs/version-3.13/whats-new.md
pstack2021: So when I read this: "Classic queues can use [the newer message store v2](https://3564b8ce.rabbitmq-website.pages.dev/blog/2024/01/11/3.13-release)", the first thing I do is click the link, I am then looking for a heading: "Message Store" or something like it to take me directly to more information this "newer v2 message store", I know the section is this one: https://3564b8ce.rabbitmq-website.pages.dev/blog/2024/01/11/3.13-release#classic-queues-storage-primer but this section in the blog is extremely long...multiple paragraphs and the punch line of what I am looking for is here: 

The per-vhost message store is still there for larger messages, but version 3.13 changes its behaviour significantly.

_For backwards compatibility queue_index_embed_msgs_below still controls whether a message is large enough to be stored in the per-vhost message store and the default is still 4kB.

It's worth mentioning that the per-vhost message store greatly improves support for fanout scenarios (when a single message is delivered to multiple classic queues), since RabbitMQ can just write the message once, rather than once per queue._

I guess what I am saying is the snippet you have in the "What's New" is ok but you have to go "the detail" in the blog to find out more about this "newer message store", the blog information needs to be clearer to get the user to find this information faster. 

Can you update the blog to do this? So define exactly what this "newer message store V2" is at the top of the blog and describe its benefits, then next, we can go on to explain briefly how classic queues store messages (or do we even need to do that), if we do perhaps we can shorten this section more.

versioned_docs/version-3.13/whats-new.md
pstack2021: How about: "Just like quorum queues, Khepri is also based on Raft protocol. As a result, the semantics of what to do when a partition occurs are well defined and not configurable."

versioned_docs/version-3.13/whats-new.md
pstack2021: or else @dumbbell do this in the "Classic Queue Performance Improvements" section in the "What's new", define exactly what this "newer message store V2" is at the top of the blog and describe its benefits and then link out to the blog. For future blogs we should use a more structured approach to the sections i.e. Title = feature name -> short opening paragraph = a couple of lines explaining clearly what this new feature is -> provide some benefits -> provide an example, then we can pull this information (well without the example) directly from the blog and into our What's New.

versioned_docs/version-3.13/whats-new.md
dumbbell: Thank you, applied!

versioned_docs/version-3.13/whats-new.md
dumbbell: I see. I don't know the change well enough to explain clearly what it is and the benefits and don't want give an incorrect summary. @essen, could you please write a short paragraph that explains what the message store v2 is and its benefits?

lhoguin: This refers/links to this blog post: https://www.rabbitmq.com/blog/2024/01/11/3.13-release

Isn't that good enough? :)

It shouldn't be "can use" but "use" though. The 3.13 changes are for all CQs. It also shouldn't say "v2" as the 3.13 changes don't fundamentally change the message store. So something like

```
Classic queues [shared message store](/blog/2024/01/11/3.13-release) brings significant performance
improvements, in terms of ...

dumbbell: Thanks @lhoguin, I applied your suggestion!

@pstack2021: What do you think?

PR #1886, opened 2024-04-18
Fix typo in clustering.md

regan-karlewicz-finsight:


michaelklishin:
This will need backporting to `versioned-docs/version-3.13` in order to go live.

PR #1891, opened 2024-04-22
Fix typo in vhosts.md

uncle-roma:
I copy-paste some commands from docs and find typo


michaelklishin:

PR #1894, opened 2024-04-24
Fix: Add double quotes in CLI command in the "Users and Authentication" section

jpoliveira08:
### **Problem Description:**
While setting up a new user following the documentation at the section "Users and Authentication" in the following [link](https://www.rabbitmq.com/docs/mqtt#authentication). I encountered an error when executing the command `rabbitmqctl set_permissions -p / mqtt-test ".*" ".*" ".*"`. The error presented was "Error (argument validation): too many arguments".

image

### **Proposed Solution:**
To fix this issue, I added double quotes around the "/" character. This change resolves the error and allows the command to be executed as expected.

### DIFF
--- a/docs/mqtt.md
+++ b/docs/mqtt.md
@@ -285,7 +285,7 @@ to the default [virtual host](./vhosts) used by this plugin:
 ```bash
 # username and password are both "mqtt-test"
 rabbitmqctl add_user mqtt-test mqtt-test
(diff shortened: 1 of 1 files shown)


jpoliveira08:

michaelklishin:
This should not be necessary. I cannot reproduce this behavior in two shells (`zsh` and `bash`). But escaping the virtual host argument also won't break anything.

PR #1899, opened 2024-04-28
Fix typo in command of migrate-mcq-to-qq.md doc

nbompetsis:
Hello folks!

Recently, I tried to perform the migration guide from the mirrored classic to quorum queues in one of my company's clusters, and I observed a minor typo in the command that imports the new schema that contains the quorum type in the queues.

`rabbitmqadmin import -V NEW_VHOST NEW_VHOST.json
`

I observed it and I decided to fix it...

Thank you in advance as well as for the migration guide, it was quite helpful!!


michaelklishin:
Honestly, I think the best approach would be to recommend `rabbitmqctl import_definitions` instead of even `rabbitmqadmin` v2.

PR #1916, opened 2024-05-15
Fix typo: wtih should be with

orvillesimba:
`compatible wtih Erlang` should be `compatible with Erlang`.


michaelklishin (approved):

michaelklishin:
Thank you!

PR #1918, opened 2024-05-16
Typo fix: overriden should be overridden

orvillesimba:
Hi it's me again 🤣, I've found this additional typo.


michaelklishin:
Thank you for your ongoing contributions to RabbitMQ docs!

PR #1949, opened 2024-06-13
fix(quorum-migration): clarify docs on quorum/classic comp

AndrewWinterman:
I was looking at the flow of the entire document with the recent change, and decided that it was actually pretty unclear. This moves the future-session-compatibility concern to its own section at the end of the migration process.

I also added mention of the`quorum_queue.property_equivalence.relaxed_checks_on_redeclaration` policy, since that's what we're going to use for our own migration.


michaelklishin:
Thank you! That's indeed a useful option that the blog post did not mention because its original version likely pre-dates RabbitMQ `3.11.16`.

PR #1956, opened 2024-06-20
Explain how to achieve read-only management user

MarcialRosales:
Explain how to achieve read-only management user. This is the proposed solution to an issue raised in rabbitmq-server https://github.com/rabbitmq/rabbitmq-server/issues/10526 .

### DIFF
--- a/versioned_docs/version-3.13/management/index.md
+++ b/versioned_docs/version-3.13/management/index.md
@@ -267,7 +267,10 @@ queues and bindings through the management
 plugin or other means.
 
 All users can only list objects within the virtual
-hosts they have any permissions for.
+hosts they have any permissions for. However, if you only want those users to have "read-only" access, 
+in order words, you want them to list all objects and view all their metrics but you do not want them to
+modify those objects (queues or streams, exchanges, bindings) in any way (consume, publish, delete or create) all you need to do is grant one permission with an empty regular expression and on any of the three actions, `read` or `write`
+or `configure`. 
 
 If access to management UI is impossible to due the lack of users
 with sufficient permissions or forgotten/incorrect permissions, [CLI tools](./cli) must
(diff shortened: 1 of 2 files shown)


michaelklishin:
I've finished this PR in https://github.com/rabbitmq/rabbitmq-website/commit/f6ad08014d21f976887c6aff6744db0020835eba and merged it manually while @MarcialRosales is away for a couple of weeks.

versioned_docs/version-3.13/management/index.md
```diff
@@ -267,7 +267,10 @@ queues and bindings through the management
 plugin or other means.
 
 All users can only list objects within the virtual
-hosts they have any permissions for.
+hosts they have any permissions for. However, if you only want those users to have "read-only" access, 
+in order words, you want them to list all objects and view all their metrics but you do not want them to
+modify those objects (queues or streams, exchanges, bindings) in any way (consume, publish, delete or create) all you need to do is grant one permission with an empty regular expression and on any of the three actions, `read` or `write`
+or `configure`.
```
pstack2021, line 274: Hey @MarcialRosales, I know this PR is merged now but please see below for a suggested rewrite, please ensure I have kept he technical meaning in this rewrite.

One comment on the first sentence "All users can only list objects within the virtual
hosts they have any permissions for.", this sentence was always there so this is not part of your updates in this PR but just mentioning it here, the previous sentence shouldn't really end with the word "for". It is fine to finish a sentence with a preposition in casual writing but not in formal business writing.

How about this:


Objects within the virtual hosts that users have permissions for can only be listed. However, if you want these users to have "read-only" access, that is, you want list all objects and view all their metrics for these users but you do not want these users to modify these objects (modify queues or streams, exchanges, or bindings) or consume, publish, delete or create these objects, then you can grant one permission with an empty regular expression on any of the three actions: `read`, `write` or `configure`."

I am not 100% sure what "grant one permission with an empty regular expression and on any of the three actions, `read` or `write`or `configure`.  actually means but if your readers/users will that as fine leave as is based on my rewrite above.
Thanks

PR #1967, opened 2024-07-09
Request from product management to remove End of Extended Commercial Support info from the Release Information topic

pstack2021:
Request from product management to remove "End of Extended Commercial Support info from the Release Information topic. @HowardTwine, can you please provide more information here. 

@michael-k or @dumbbell can one of you please complete further updates to delete the dates (computation) and the display of the column, my commit is only deleting the header, many thanks.

### DIFF
--- a/src/components/RabbitMQServerReleaseInfo/index.js
+++ b/src/components/RabbitMQServerReleaseInfo/index.js
@@ -281,12 +281,6 @@ export function RabbitMQServerReleaseInfoTable() {
             "release-eos-community"
           ].join(' ')}>End of Community Support</div>
 
-          <div className={[
-            "release-info-header",
-            "release-eos",
-            "release-eos-commercial"
-          ].join(' ')}>End of Extended Commercial Support</div>
-
           {rows}
         </div>
       </div>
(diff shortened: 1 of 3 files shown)


src/components/RabbitMQServerReleaseInfo/index.js
```diff
@@ -281,12 +281,6 @@ export function RabbitMQServerReleaseInfoTable() {
             "release-eos-community"
           ].join(' ')}>End of Community Support</div>
 
-          <div className={[
-            "release-info-header",
-            "release-eos",
-            "release-eos-commercial"
-          ].join(' ')}>End of Extended Commercial Support</div>
-
           {rows}
```
pstack2021, line 284: The reason for this change is to remove the potentially confusing information regarding the fixed term extended commercial support in this table. The Broadcom approach is more flexible in that we are taking customer requirements for extended commercial support into account on a case by case basis.

michaelklishin:
FTR, this was reverted with certain changes in https://github.com/rabbitmq/rabbitmq-website/pull/1998.

PR #1971, opened 2024-07-11
Added architecture filter to deb examples.

Dawnthorn:
On ubuntu 22.04 after following these instructions, I was getting: 
```
Err:11 https://ppa1.novemberain.com/rabbitmq/rabbitmq-server/deb/ubuntu jammy/main i386 Packages
  404  Not Found [IP: XXXX 443]
```
and
```
E: Failed to fetch https://ppa1.novemberain.com/rabbitmq/rabbitmq-erlang/deb/ubuntu/dists/jammy/main/binary-i386/Packages  404  Not Found [IP: XXXXX 443]
```

Someone mentioned that in discussions:
https://github.com/rabbitmq/rabbitmq-server/discussions/8934

The answer was to use a 64-bit machine and I am using a 64 bit machine. I suspect this happens on Ubuntu if you have it setup to bring in 32 bit libraries for some things, but I've not checked out if this is true.

I think that since the PPA only provides AMD64 packages, adding the filter shouldn't mess anyone up.

### DIFF
--- a/docs/install-debian.md
+++ b/docs/install-debian.md
@@ -183,20 +183,20 @@ curl -1sLf https://github.com/rabbitmq/signing-keys/releases/download/3.0/clouds
 sudo tee /etc/apt/sources.list.d/rabbitmq.list <<EOF
 ## Provides modern Erlang/OTP releases
 ##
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you, this is indeed a good idea, at least for now.

PR #1976, opened 2024-07-15
MQTT plugin authenticate using client_id from client certificate

MarcialRosales:
Implements feature request https://github.com/rabbitmq/rabbitmq-server/issues/9323

### DIFF
--- a/docs/mqtt.md
+++ b/docs/mqtt.md
@@ -596,11 +596,11 @@ global parameter and so takes precedence over it.
 
 #### Use the client_id from client certificate to authenticate 
 
-You can configure RabbitMQ to ensure the `client_id` received over the MQTT connection matches the `client_id` found in the client's certificate. If they match, RabbitMQ proceeds with the user's authentication by passing the user's identity along with the `client_id` to the configured authentication backend(s). Some authentication backends, like the `rabbitmq_auth_backend_http`, may use `client_id` credential in addition to `username` to make authentication and/or authorization decisions. 
+You can configure RabbitMQ to ensure the `client_id` received over the MQTT connection matches the `client_id` found in the client's certificate. If they match, RabbitMQ proceeds with the user's authentication by passing the user's identity along with the `client_id` to the configured authentication backend(s). Some authentication backends, like the `rabbitmq_auth_backend_http`, may use `client_id` credential in addition to `username` to make authentication and/or authorization decisions. If the `client_id`(s) did not match, RabbitMQ closes the connection with the reason code `2`, meaning, `the client identifier is not allowed by the server`.
 
 First, you must configure where RabbitMQ extracts the `client_id` from the certificate by setting the configuration variable `mqtt.ssl_cert_client_id_from`. The acceptable options are:
 - `distinguished_name`, this is the DN, or distinguished name, of the certificate
-- `subject_alternative_name` or `subject_alt_name`, here the `client_id` is specified in certificate's extension. You can specify further alternative names which can be of several types.
+- `subject_alternative_name`, here the `client_id` is specified in certificate's extension. You can specify further alternative names which can be of several types.
 
 If you set `mqtt.ssl_cert_client_id_from` to `subject_alternative_name`, you can configure the type of alternative name via the `mqtt.ssl_cert_client_id_san_type` configuration variable. If you do not set it, its default value is `dns`. The acceptable options are:
 - `dns`
--- a/docs/mqtt.md
+++ b/docs/mqtt.md
@@ -594,7 +594,7 @@ to reconnect to switch to a new virtual host.
 is considered more specific than the port-to-vhost mapping with the `mqtt_port_to_vhost_mapping`
 global parameter and so takes precedence over it.
 
-#### Use client_id from client certificate to authenticate 
+#### Use the client_id from client certificate to authenticate 
 
 By the default, the plugin passes the `client_id` as credential to the configured authentication backends. However, it is possible to extract the `client_id` from the certificate. 
 
--- a/docs/mqtt.md
+++ b/docs/mqtt.md
@@ -598,12 +598,12 @@ global parameter and so takes precedence over it.
 
 By the default, the plugin passes the `client_id` as credential to the configured authentication backends. However, it is possible to extract the `client_id` from the certificate. 
 
-First, you must configure which certificate's attribute has the `client_id` by setting the configuration variable `mqtt.ssl_cert_client_id_from`. The acceptable values are:
+First, you must configure where RabbitMQ extracts the `client_id` from the certificate by setting the configuration variable `mqtt.ssl_cert_client_id_from`. The acceptable options are:
 - `distinguished_name`, this is the DN, or distinguished name, of the certificate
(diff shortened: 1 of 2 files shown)


docs/mqtt.md
```diff
@@ -594,6 +594,31 @@ to reconnect to switch to a new virtual host.
 is considered more specific than the port-to-vhost mapping with the `mqtt_port_to_vhost_mapping`
 global parameter and so takes precedence over it.
 
+#### Use client_id from client certificate to authenticate 
+
```
pstack2021, line 598: Please change the title to: Use the client_id from the Client Certificate to Authenticate

I also added capitalization to some of the words in the heading, please check other sub section titles in OSS RabbitMQ topics, if they use capitalisation for 2nd/3rd words in titles then please keep it consistent here, if they don't then you can leave as is, thanks.

docs/mqtt.md
```diff
@@ -594,6 +594,31 @@ to reconnect to switch to a new virtual host.
...
+By the default, the plugin passes the `client_id` it received from the MQTT connection to the configured authentication backends as a credential. However, it is possible to extract the `client_id` from the certificate in a very similar way as we extract the username. 
+
```
pstack2021, line 600: Perhaps rephrase the 1st sentence to: By the default, the plugin passes the `client_id` as a credential to the configured authentication backends.

Do they need to know this < which it received from the MQTT connection> if not I would leave it out?

pstack2021: For the 2nd sentence how about: "However, it is possible to extract the `client_id` from the certificate using the same way the username is extracted. "

Is is the same way or similiar? similar is a bit vague?

MarcialRosales: You are right, we may leave that detail out. Thanks!

docs/mqtt.md
```diff
@@ -594,6 +594,31 @@ to reconnect to switch to a new virtual host.
...
+First, you must configure which certificate's attribute has the `client_id` by setting the configuration variable `mqtt.ssl_cert_client_id_from`. The allowed values are:
```
pstack2021, line 601: What does this mean: "ou must configure which certificate's attribute has the `client_id` " does the user need to search which one is it? 

I would say "The acceptable values are:"

docs/mqtt.md
```diff
@@ -594,6 +594,31 @@ to reconnect to switch to a new virtual host.
...
+- `distinguished_name`, this is the full DN, or distinguished name, of the certificate
```
pstack2021, line 602: Here I would just say, "this is the full distinguished name, of the certificate." is there a reason for saying "the full DN"?? if not, I would leave it out

docs/mqtt.md
MarcialRosales: how about: `you must configure where RabbitMQ extract the client_id from the certificate?`

docs/mqtt.md
```diff
@@ -594,6 +594,31 @@ to reconnect to switch to a new virtual host.
...
+- `common_name`, this is just the CN part of the distinguished name
+- `subject_alternative_name` or `subject_alt_name`, here the `client_id` is specified in certificate's extension which allows to specify further alternative names which can be of several types.
```
pstack2021, line 604: again I would rephrase: ", in these X, the `client_id` is specified in the certificate's extension. You can specify further alternative names which can be several types."

docs/mqtt.md
MarcialRosales: umm. but the outer title at level 3 does not have that type of capitalization ... `Authentication with TLS/x509 client certificates` ... ??

docs/mqtt.md
```diff
@@ -594,6 +594,31 @@ to reconnect to switch to a new virtual host.
...
+
+If you set `mqtt.ssl_cert_client_id_from` to `subject_alternative_name` or `subject_alt_name`, you should configure `mqtt.ssl_cert_client_id_san_type` which contains the `client_id`, otherwise it defaults to `dns`, its type which can be:
```
pstack2021, line 606: ?? "you should" or "you must" in the 1st sentence? Do they have to do it, "you should" is too vague, if you are going to use "should", I would additionally explaining why they should, thanks

pstack2021: its type which can be:

??? should that be a new sentence written as "It's type can be: "

docs/mqtt.md
```diff
@@ -594,6 +594,31 @@ to reconnect to switch to a new virtual host.
...
+- `dns`
+- `ip` 
+- `email` 
+- `uri`
+- `other_name`
+
+And if there can be more than alternative name of the type specified, you can configure which one to choose via the `mqtt.ssl_cert_client_id_san_index` configuration variable. By default, it always pick the first name, i.e. `mqtt.ssl_cert_client_id_san_index = 0`.
```
pstack2021, line 613: can't start a sentence with "And" so how about "If there can be additional alternative names of the type that is specified,......

docs/mqtt.md
```diff
@@ -594,6 +594,31 @@ to reconnect to switch to a new virtual host.
...
+
+Here is an example where the username is extracted from the certificate's `common_name` and the `client_id` from the first SAN -subject alternative name- of type uri:
+
```
pstack2021, line 616: I would say "and the `client_id` is extracted....

versioned_docs/version-3.13/mqtt.md
```diff
@@ -594,6 +594,31 @@ to reconnect to switch to a new virtual host.
 is considered more specific than the port-to-vhost mapping with the `mqtt_port_to_vhost_mapping`
 global parameter and so takes precedence over it.
 
+#### Use client_id from client certificate to authenticate 
+
```
pstack2021, line 598: Apply same comments as above here

docs/mqtt.md
MarcialRosales: I chose `should` because if you do not set it, it is defaulted to `dns`. If that is what the user wants, then the user does not need to configure it.

docs/mqtt.md
MarcialRosales: I keep forgetting that rule !! :(

docs/mqtt.md
MarcialRosales: i leave out "the same way " .. to avoid any confusion.

docs/mqtt.md
pstack2021: ok if the others don't have it, leave as it but include the additional words I suggested, thanks

docs/mqtt.md
pstack2021: then I would say "If you want `mqtt.ssl_cert_client_id_from` to  be set to`subject_alternative_name` or `subject_alt_name`, then you must.....

docs/mqtt.md
pstack2021: No problem :)

docs/mqtt.md
MarcialRosales: At the end, i have gone for "you can configure ".. if the user wants to use `dns` as the SAN type, it is not required to set it `mqtt.ssl_cert_client_id_san_type` to `dns` as that is the default behaviour. Are ok ?

PR #1977, opened 2024-07-16
docs/install-debian: add ERL_MAX_PORTS check

mitchdz:
It would be nice to have docs on how to check on how many ports rabbitmq can open, if ERL_MAX_PORTS differs from LimitNOFILE. You can find this command from[0].

[0] - https://www.erlang.org/docs/26/man/erl#+Q


michaelklishin:
I'm afraid we cannot accept this PR as is. It misplaces the recommendation (`ERL_MAX_PORTS` is covered in a different guide and in the Debian guide alone, there are several places where such recommendation would be relevant), does not update the RPM guide, only updates the 4.x version of the docs, and spells RabbitMQ as "rabbitmq" while the rest of the documentation does not (unless referring to CLI tool names).

I will push an alternative version.

mitchdz:
Thanks for working on an alternative version Michael. This is my first time contributing to these docs so I wasn't sure the best place to put it. Will note down the other comments for any future contributions :)

PR #1981, opened 2024-07-17
oauth2 scope documentation/typo fix

m1schka-bdr:
match the scope code with the written documentation


m1schka-bdr:
thinking about it again, it could also be that it was correct (at this position) but then it´s wrong at line 367
let me know

michaelklishin:
It should be `x-*` for exchange name, since the section is dedicated to topic permissions (which only apply to topic exchanges).

Thank you.

m1schka-bdr:
but then, as I corrected myself, everything should start with `x-`. should I provide another PR or do you want to leave it as it is right now?

m1schka-bdr:
nevermind, I just saw you corrected it in a follow up commit. thx

PR #1983, opened 2024-07-22
Doc regarding reconciliation

SimonUnge:


SimonUnge:
@michaelklishin Rough draft but wonder if this aligns with your Doc vision!

michaelklishin:
This only updates the 4.0 version (accessible at `/docs/next/quorumq-queues/` and not the 3.13 one under `versioned_docs`.

SimonUnge:
Aha, ok will fix that.

PR #1993, opened 2024-07-31
added documentation for `check_if_cluster_has_classic_queue_mirroring_policy` 

markus812498:
Found rabbitmqcli command that wasn't documented, so added documentation around its usage.
introduced in https://github.com/rabbitmq/rabbitmq-server/pull/9869 - can add the others too.

### DIFF
--- a/docs/man/rabbitmqctl.8.md
+++ b/docs/man/rabbitmqctl.8.md
@@ -1487,6 +1487,17 @@ fanout
           </div>
         </dd>
         <dt >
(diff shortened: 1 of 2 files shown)


markus812498:
I just saw this: https://www.rabbitmq.com/docs/man
feel free to decline if I should PR this elsewhere.

michaelklishin:
The source of the page is in the server repo under docs. It uses ntroff, the man page format. There are  Make and Bazel targets that generate the HTML versions.

michaelklishin:
https://github.com/rabbitmq/rabbitmq-server/pull/12005.

PR #1994, opened 2024-08-05
streaming tutorial_2 for python and rust

DanielePalaia:
Tutorial two for RabbitMQ streams for Python and Rust clients.

It fixes a small thing on the Python tutorial1 as we don't need to install the type_extension dependency.


DanielePalaia:
Rust client tutorial_2 hidden till this bug in the Client (necessary for the tutorial): https://github.com/rabbitmq/rabbitmq-stream-rust-client/issues/222 get fixed

acogoluegnes (approved):

PR #1996, opened 2024-08-05
Add AMQP brokers

ansd:
Clarify that Apache Qpid Broker-J also supports AMQP 0.9.1 as described in https://github.com/apache/qpid-broker-j?tab=readme-ov-file#qpid-broker-j


ansd:
Thanks @johanrhodin !

PR #2003, opened 2024-08-08
Fix typo on firehose tracing page

jimmielyon:


michaelklishin:
Thank you for updating the 3.13.x version, too.

PR #2007, opened 2024-08-12
Modified the links in feature matrix table

jinskjoy:
The link of Non-durable Queues and Exclusivity are linked to another page. It's confusing to the reader and until the user scrolls down to the end of table, the section with short descriptions are not visible.

The short description area has links to detailed description pages.


michaelklishin:
Thank you for taking the time to contribute.

We link to other guides all the time. The local sections you have linked to in turn link to the queues guide, so I don't see how this change makes anything different.

This table has been using external links for certain features for a few years now and FWIW this hasn't been a commonly suggested change.

PR #2012, opened 2024-08-13
Client options can't be set globally in rabbitmq.conf

AntonSigur:
amqp_client.ssl_options are not available to be set in the rabbitmq.conf file - only available in the advance.conf file.


michaelklishin:
@AntonSigur read what you are removing carefully. The part talks about `rabbitmq.config`, the original configuration file, not `rabbitmq.conf`.

michaelklishin:
I have pushed a similar set of updates. There are no reasons to mention `rabbitmq.config` in general, and the example was missing a trailing dot, which is not optional in `advanced.config`.

Thank you for taking the time to contribute.

PR #2017, opened 2024-08-15
Document SASL mechanism ANONYMOUS

ansd:
RabbitMQ 4.0 docs for https://github.com/rabbitmq/rabbitmq-server/pull/11999


michaelklishin:
Merged manually with some edits. Most importantly, the current example was from the `rabbitq.config` (the Erlang term format) era, I have updated it to use the format in `rabbitmq.conf`.

PR #2040, opened 2024-09-03
Correct description of OAuth2 preferred username claims config

iinuwa:
Since 3.10/[rabbitmq/rabbitmq-server#7458](https://github.com/rabbitmq/rabbitmq-server/pull/7458), preferred_username_claims has accepted a list of claim names rather than a comma-separated string.


iinuwa:

PR #2041, opened 2024-09-03
Replace preferred_username_claims advanced.config example with modern…

iinuwa:
… config example

Since the project recommends using the modern config example, we should use the modern config in the docs wherever possible.


iinuwa:

PR #2043, opened 2024-09-04
Avoid invalid admonition syntax

ansd:
As described in https://docusaurus.io/docs/markdown-features/admonitions#usage-with-prettier
> If you use Prettier to format your Markdown files, Prettier might auto-format your code to invalid admonition syntax.
> To avoid this problem, add empty lines around the starting and ending directives. This is also why the examples we
> show here all have empty lines around the content.


michaelklishin:
This syntax without the empty lines is not invalid, Docusaurus' own docs use it. 

We don't use empty lines in admonitions in every other guide. I guess we also do not format them with Prettier so it becomes a matter of taste.

PR #2044, opened 2024-09-04
OAuth 2 guide updates

MarcialRosales:
Note: Entra/Azure example is still using v1 url for the oauth_provider_url and v2 for the jwks_url. At this time, it is not possible to use v2 also in the oauth_provider_url until https://github.com/rabbitmq/rabbitmq-server/issues/12235 is completed.

I have also noticed that the latest Entra/Azure's user interface differs a bit from the screenshots attached to the example. Once the aforementioned PR is completed, I will update the screenshots on a separate docs PR


docs/oauth2-examples-azure/index.md
```diff
@@ -157,27 +157,21 @@ Update it with the following values:
...
+auth_backends.2 = rabbit_auth_backend_internal
+
+management.oauth_enabled = true
+management.oauth_client_id = <YOUR_AZURE_AD_TENANT_ID>
+management.oauth_scopes = openid profile rabbitmq.tag:administrator
+management.oauth_provider_url = https://login.microsoftonline.com/<YOUR_AZURE_AD_TENANT_ID>
+
+auth_oauth2.resource_server_id = <YOUR_AZURE_AD_TENANT_ID>
+auth_oauth2.additional_scopes_key = roles
+auth_oauth2.jwks_url = <<"PUT YOUR AZURE AD JWKS URI VALUE">>
```
michaelklishin, line 171: Using a binary here may confuse some people. `rabbitmq.conf` does not accept the binary syntax, even though a mapping type can be used to produce a binary out of a value.

MarcialRosales: It was unfortunate that i chose << " " >> :) I will use <PUT YOUR AZURE AD JWKS URI VALUE> instead. I did not mean erlang binary value.

michaelklishin:
I have fixed the build. @MarcialRosales we should probably do this renaming for the 3.13.x version, too?

MarcialRosales:
@michaelklishin , yes, once I finished doing all the changes in the docs , I apply those changes to 3.13.x too.

PR #2059, opened 2024-09-10
Start documenting AMQP client libraries

acogoluegnes:


michaelklishin:
@acogoluegnes do you intend this guide to cover all clients? Given the size of the Java and .NET AMQP 0-9-1 guides, I suspect it can grow quite large.

PR #2068, opened 2024-09-18
Refresh upgrade-related documentation

mkuratczyk:
* split the `upgrade` page into multiple pages
* each upgrade strategy has a dedicated page
* right-hand side table of contents should outline the main steps
* remove references to full-stop upgrades - there has been no need for a full stop upgrade since feature flags were introduced
* remove other notes that are no longer relevant

### DIFF
--- a/docs/blue-green-upgrade.md
+++ b/docs/blue-green-upgrade.md
@@ -19,10 +19,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 
-Blue-green deployment is an [upgrade strategy](./upgrade) that is based on the idea of setting up
-a second RabbitMQ cluster (the "green" one) next to the current production
-cluster (the "blue" one). Applications are then switched to the "green"
-cluster. When that migration is done, the "blue" cluster is decommissioned (shut down).
+Blue-green deployment is a migrtion technique that can also be used as an [upgrade strategy](./upgrade).
+The main idea is to set up a new environment (the "green" one) and switch to it
+when it is ready. Technically nothing is upgraded - the application just switch
+to a different environment, which might be using a different version, but can
+also differ in other aspects. For example, the same approach can be used
+to migrate to new hardware, while keeping the same version of RabbitMQ.
+
+When that migration is done, the old ("blue") cluster is decommissioned (shut down, deleted).
 To simplify the switch, [federated queues](./federated-queues)
 can be used to transfer enqueued messages from the "blue" to the "green" cluster.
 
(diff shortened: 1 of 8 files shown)


docs/blue-green-upgrade.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Upgrading RabbitMQ Using Blue-Green Deployment Strategy
+title: Blue-Green Deployment
```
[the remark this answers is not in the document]
mkuratczyk: I've pushed a commit that uses "blue-green deployment" more consistently

docs/grow-then-shrink-upgrade.md
```diff
@@ -0,0 +1,131 @@
...
+This approach may seem like one that strikes a good balance between the relative simplicity of
+in-place upgrades and the safety of Blue-Green deployment upgrades. However, in practice this
+strategy has comparable characteristics to the in-place upgrade option:
+
+ * Newly added nodes may affect the existing cluster state
+ * Replicas will migrate between nodes during the upgrade process
+
+In addition, this approach has its own unique potential risks:
+
+ * Node identities change during the upgrade process, which can affect [historical monitoring data](./monitoring/)
```
dumbbell, line 61: Unwanted trailing slash at the end of the URL.

docs/upgrade.md
```diff
@@ -107,404 +102,195 @@ nodes, and is only safe with important precautions. Therefore, it is [highly rec
...
-### Release Series Upgradeability with Rolling Upgrades
+<details>
+  <summary>Release Series Upgradeability</summary>
 
-Current release series upgrade compatibility with **rolling** upgrade:
+The following shows the supported upgrade paths.
 
 | From     | To     | Notes                                                         |
 |----------|--------|---------------------------------------------------------------|
+| 3.13.x   | 4.0.x  | Some feature flags **must** be enabled **before** the upgrade |
```
dumbbell, line 140: Why not say "All stable feature flags must be enable before (…)"? As opposed to "some" because it gives the impression that the user is responsible for figuring out which one are to be enabled and if they make a mistake, their cluster will crash and two kittens 100 km from there will die.

dumbbell (changes requested):

docs/grow-then-shrink-upgrade.md
```diff
@@ -0,0 +1,131 @@
+---
+title: Grow-then-Shrink Upgrade
+displayed_sidebar: docsSidebar
```
dumbbell, line 3: It shouldn’t be necessary to specify the `displayed_sidebar` because the document is listed in the sidebar already.

dumbbell (changes requested):

PR #2081, opened 2024-09-25
Fix links

ansd:
see https://github.com/rabbitmq/rabbitmq-server/discussions/12377#discussioncomment-10753607


lukebakken (commented):
👍 🌮 thanks!

PR #2083, opened 2024-09-27
fix typo in prometheus doc

elelayan:
should be exchange instead of queue


elelayan:
“@pivotal-cla This is an Obvious Fix

michaelklishin:
@elelayan you have updated the 4.1 version of the site. I have applied the same change to the 4.0 (current) version. It should be live in a few minutes.

Thank you for contributing to RabbitMQ docs.

elelayan:
thanks for the quick merge

PR #2085, opened 2024-09-30
link to `cluster-operator` examples

peterbecich:
Specifically, the example to modify the default resource requests and limits


michaelklishin:
Thank you.

PR #2093, opened 2024-10-04
Remove space from :::important tag in dynamic shovel 1.0 html section

MrRoss256:
::: important had a rogue space causing not to render properly. Copied from same edit earlier in the 0.9.1 table.  I have not tested this change but there is a working version in the same document and confident this is a valid change.


michaelklishin:
This only updates the 4.1 version of the docs.

michaelklishin:
@MrRoss256 thank you for bringing this up. I have updated the 4.0 and 3.13 versions, the changes should be live by now.

PR #2098, opened 2024-10-14
fixed duplicate IDs in Connections Docs

IPdotSetAF:


IPdotSetAF:

michaelklishin:
The right thing to do here is to delete the duplicate section entitled TLS.

PR #2102, opened 2024-10-17
update documentation

Gsantomaggio:


acogoluegnes (approved):

michaelklishin:
@Gsantomaggio @acogoluegnes was this meant to only update the 4.1 version of the docs? Should we at the very least update the `versioned_docs/*` variant for 4.0?

acogoluegnes:
The change was only for the client library page, which is not versioned IIUC.

PR #2105, opened 2024-10-23
Document `x-cc` message annotation

ansd:
Document the new `x-cc` message annotation as introduced in RabbitMQ 4.1 via https://github.com/rabbitmq/rabbitmq-server/pull/12559

Rewrite AMQP 0.9.1 docs since they didn't make sense.


michaelklishin (approved):

michaelklishin:
This wording for AMQP 0-9-1 is definitely an improvement.

PR #2084, opened 2024-09-27
4.1: Deprecate two OAuth 2 settings

MarcialRosales:
Implements these two features:
- https://github.com/rabbitmq/rabbitmq-server/issues/12239
- https://github.com/rabbitmq/rabbitmq-server/issues/12237

Only for 4.1.x

**This PR should only be merged after this [PR](https://github.com/rabbitmq/rabbitmq-server/pull/12399) has been merged.** and this PR has been rebased too.

PR #2106, opened 2024-10-23
Fix the link to rabbitmq.conf link to point to the file the does exist

jasonholmberg:
This link to `rabbitmq.conf` at this page is broken:

https://www.rabbitmq.com/docs/oauth2-examples-okta#configure-rabbitmq-to-use-okta-as-oauth-20-authentication-backend

This PR attempts to fix that broken link by point to the rabbitmq.conf.tmpl file that does exist in that location.


michaelklishin:
For 4.1/`main`, it was addressed as a drive-by change in #2084.

michaelklishin:
@jasonholmberg thank you!

PR #2108, opened 2024-10-24
Update install-homebrew.md

frompinewood:
Fixed markdown for bash highlighting


frompinewood:
This is a typo correction and should not require a CLA signature.

michaelklishin:
This only updates the `main`/4.1 version and not the versions under `versioned_docs`.

michaelklishin:

PR #2116, opened 2024-11-01
[TNZDOC-342] Language edits and link fixes

RichardJJG:


michaelklishin:
@RichardJJG thank you. Note that this only updates the 4.1 version of the docs. I will clarify something with other engineers and backport what should be backported on your behalf. Right now it looks like all the changes can be backported but I am a bit less certain about the AMQP 1.0 part which has been updated recently specifically for 4.1.0.

michaelklishin:
`4.0.x` and `3.13.x` versions are under `./versioned_docs`.

PR #2126, opened 2024-11-09
Fix typos in Quorum Queue -> Resource Use section

tarikdemirci:


michaelklishin:
Thank you for taking the time. This only update the 4.0.x version but not 4.1 (main) and 3.13, I will update the other two on your behalf.

PR #2134, opened 2024-11-20
Fix topology operator docs: sync period

Zerpet:
Documented default behaviour is incorrect. Based on the docs from
controller-runtime library, the default sync period is 10 hours. Adding
how to disable this mechanism, as it may not be trivial to deduce.


Zerpet:
Fixes https://github.com/rabbitmq/messaging-topology-operator/issues/864

michaelklishin (approved):

PR #2141, opened 2024-12-13
Add a hint on Debian package conflicts for 3.13

bekoev:
To the `3.13.x` version of the site add a hint for the case of conflicts with Erlang versions, so that fellow users do not add new issues or questions facing problems with the installation instruction.

I have not run tests, and also hope the change is trivial enough to not sign the contributor agreement.

### DIFF
--- a/versioned_docs/version-3.13/install-debian.md
+++ b/versioned_docs/version-3.13/install-debian.md
@@ -154,6 +154,8 @@ and <a href="#apt-cloudsmith-erlang">on Cloudsmith.io</a>.
 
 Below is a shell snippet that performs those steps.
 
+Note: by default, the snippet installs the latest version of Erlang available in the apt repository. If you face vesion conflicts, note the section on [package pinning](#apt-pinning).
+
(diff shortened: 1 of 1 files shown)


bekoev:

michaelklishin:
Thank you, I had something else in mind. Adding version suffixes that would install `26.2.5.6` or something with a comment would be more convenient.

bekoev:
Would you prefer that I add `sudo apt-get install -y erlang-base=1:26.2.5.6-1 \ ...` to all the snippets? Which testing scope would be sufficient in that case (the website page view, testing updated instructions using Ubuntu/Debian versions)?

michaelklishin:
https://github.com/rabbitmq/rabbitmq-website/commit/3ef4d4ca4afac225c5aec054981be20aab611677

michaelklishin:
Well, I have rendered the page locally to compare formatting and tested this script with `26.2.5.6` and `27.2` versions on an `amd64` Debian Bookworm machine.

PR #2144, opened 2024-12-18
Add EEF EPMD Exposure Blog Reference

maennchen:
Add blog entry with link to https://erlef.org/blog/eef/epmd-public-exposure

### DIFF
--- a/blog/2024-12-18-epmd-public-exposure/index.md
+++ b/blog/2024-12-18-epmd-public-exposure/index.md
@@ -1,0 +1,13 @@
+---
+title: "Exposed EPMD: A Hidden Security Risk"
+tags: ["Security"]
+authors: [eef]
+---
+
+## Exposed EPMD: A Hidden Security Risk
+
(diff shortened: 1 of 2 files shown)


michaelklishin:
@maennchen thank you. I cannot promise that we will publish the blog post but we certainly can adjust our documentation to communicate the idea and link to the EEF blog post.

maennchen:
@michaelklishin I thought it would be a good idea to publish something in the RabbitMQ ecosystem since not every RabbitMQ user is also well-versed with Erlang and it would be good to reach this audience as well. Especially since the nr. of affected RabbitMQ users is particularly high.

If it's in the shape of this exact post does not matter.

In any case: If the EEF security WG / myself can be of help, let me know :)

michaelklishin:
@maennchen I have merged an edited version of the post. Hopefully the general idea is intact, and the goal of this post is arguably promote the EEF blog post which goes into a lot more detail.

Thank you for contributing. I will take a look at our docs and if there's something worth adding (from the EEF blog post), update the Networking guide accodingly.

michaelklishin:
@maennchen I have also updated our Networking guide, even though `epmd`-related topics specifically were already covered.

If you have further suggestions, let us know or submit a PR (please don't forget the `versions_docs` versions).

maennchen:
@michaelklishin Awesome! I'll take some time to check the new docs.

PR #2160, opened 2025-01-07
Update consumers.md exclusivity

25schmekles:
Note that exclusive consumers is a Classic Queues only feature to reflect the same info on the feature matrix.


michaelklishin:
This only updates the 4.1 version of the docs. I will finish it.

PR #2163, opened 2025-01-10
Support forward proxy in oauth2

MarcialRosales:
Add commercial-only setting that supports explicit forward proxy and an example that demonstrates it.

Feature meant for 4.1.x.


michaelklishin:
@RichardJJG sorry, I was going to try a styling experiment to make this Tanzu-specific feature stand out but fair enough, since it is a 4.1.x series feature, merging it early and correcting later is a perfectly feasible approach :)

PR #2164, opened 2025-01-10
correct stream bloom filter example

AndrewWinterman:
I noticed the example for setting the x-stream-filter-size-bytes didn't actually set the value, so i updated it to do so.


michaelklishin:
Thank you. 

Note that this updates the `main` (4.1.0) version of the docs. I will finish this PR off on your behalf.

michaelklishin:
@AndrewWinterman thank you for contributing to RabbitMQ!

PR #2169, opened 2025-01-20
Fixed misspell: 'Deadl-lettering' -> 'Dead-lettering'

ollie-bud:


michaelklishin:
This only updates the `4.0.x` version of the site.

We will apply it to 4.1 and 3.13.x on your behalf if necessary.

PR #2174, opened 2025-01-27
fix: add note about installing from opentofu/terraform

PurpleBooth:
Encountered this snafu this morning. More [information here](https://github.com/bitnami/charts/issues/30582)


tirelibirefe:
where is the information?

michaelklishin:
@tirelibirefe [Installation with Terraform and OpenTofu](https://www.rabbitmq.com/kubernetes/operator/install-operator#installation-with-terraform-and-opentofu)

PR #2189, opened 2025-02-11
New commercial offerings and support page

skwokz:
Redesigned the /contact page. Combined the commercial-offering and contact md pages into this.


MirahImage (changes requested):
Rather than re-writing this in HTML embedded inside javascript, we would prefer you continue to format your changes in markdown, so that they can be rendered using our existing tooling. Thanks.

MirahImage (approved):

MirahImage:
@HowardTwine did you have some textual modifications you'd like?

MirahImage:
I now understand why the choice has been made. It's replicating some of the formatting from the index.js page, which is also rendered the same way.

MirahImage:
@lukebakken if you know how to produce that formatting using pure md, I think that would be an appreciated contribution, but it does appear we do the same thing ourselves in the index page.

HowardTwine:
Can we add the following to the list of exclusive enterprise features - 

Intra-cluster compression 
Reduce ingress or egress costs as well as network efficiency through the compression of network data between cluster nodes.

ansd:
> Can we add the following to the list of exclusive enterprise features -
Intra-cluster compression
Reduce ingress or egress costs as well as network efficiency through the compression of network data between cluster nodes.

@HowardTwine We should also add AMQP over WebSocket and audit logging to this list.

HowardTwine:
> > Can we add the following to the list of exclusive enterprise features -
> > Intra-cluster compression
> > Reduce ingress or egress costs as well as network efficiency through the compression of network data between cluster nodes.
> 
> @HowardTwine We should also add AMQP over WebSocket and audit logging to this list.

Audit Logger for sure but we should hold off with AMQP 1.0 over websockets until it is released (along with Scheduled and Delayed messaging).

@skwokz can you also please add the following to the Tanzu exclusive features for RabbitMQ
"Audit Logger
Output broker wide events to disk in either JSON or human readable text."

michaelklishin:
I can add those features as part of the final review.

michaelklishin (approved):

PR #2190, opened 2025-02-11
Added clarifications to the policy priorities documentation

pvladMQ:
Added clarification to the policy priorities : 

"Higher values indicate higher priority (e.g. a policy with priority 10 will overrule a policy with priority 8). " 

Some users have complained about this detail not being specific enough, leading to confusions.


MirahImage (changes requested):
This change would make more sense if added to the previous important notice (lines 354-357, which directly addresses this subject).

PR #2191, opened 2025-02-11
Add explanations to Policy Priorities

pvladMQ:
Added additional explanation to Policy Priorities: 

"Higher values indicate higher priority (e.g. a policy with priority 10 will overrule a policy with priority 8)."


pvladMQ:
@MirahImage updated according to your previous comments in #2190

MirahImage (approved):

michaelklishin:
We need to update 4.1 and 3.13 versions of the same guide. I'll do it.

PR #2202, opened 2025-03-03
More updates to Release Information and support pages

skwokz:
**Home:** 
Updated “Licensing” section and moved to the top
Added “Enterprise” tab with email call to action

**Release Information:**
Consolidated various copy on the page
Added inline link to Commercial section on Contact
Added banner to link to Commercial section on Contact
Changed table header link to go to Comercial section on Contact page

**Support:** 
Added a few new partners, updated link of one of the existing partners

**Global**
Topbar: Changed link it global bar to "Get long term support and advanced enterprise features” from version notes
Footer: Changed and updated links from the "Reach out to the RabbitMQ team” section to a “Need Help?” section
Added UTM tags to link leading to /contact#tanzu-rabbitmq


michaelklishin:
@skwokz this is quite a change to navigation. I don't think we want to make the entire website into one massive Tanzu RabbitMQ ad.

I'm clarifying with the rest of the team whether this was discussed before.

michaelklishin:
It turns out the navigation changes are in the footer, not the top-level navigation.

The "version notes" (latest release announcements) will be back together with the new release in about a month.

michaelklishin:
@skwokz I'm afraid we cannot accept these edits.

Some information provided in this PR is factually incorrect, besides that, these changes more or less remove all community support sources and OSS RabbitMQ information in favor of commercial support options and Tanzu RabbitMQ sprinkled all over the place.

I don't feel strongly about the column in the footer but why would it start not with "Learn About RabbitMQ" but straight with commercial support?

michaelklishin:
The use of the temporary (collapsable) banner for such promos is fine but when a new version comes out, it will be replaced for at least a period of time.

The changes to the Release Information page are acceptable as long as the release series table remains the central piece and not a yet another place to put some Tanzu RabbitMQ links (we already have some on that short page). I have ported over a few of these changes.

release-information/index.md
```diff
@@ -30,43 +30,23 @@ by community or extended commercial support and what release is coming next.
...
-support** to avoid confusion with the VMware [terminology related to the support lifecycle policy](https://tanzu.vmware.com/support/lifecycle_policy).
+For [regularly contributing users](https://github.com/rabbitmq/rabbitmq-server/blob/main/COMMUNITY_SUPPORT.md), community support is provided via [community resources](/contact). Refer to the previous table for end dates for this support. [Users who do not contribute](https://github.com/rabbitmq/rabbitmq-server/blob/main/COMMUNITY_SUPPORT.md) or do not hold a [valid VMware Tanzu commercial license](/contact?utm_source=rmq_release-information_text&utm_medium=rmq_website&utm_campaign=tanzu#tanzu-rabbitmq) can use community resources for questions but are not entitled to any support from the RabbitMQ Team at Tanzu.
 
-### Community Support
-
-:::important
-From June 1st, 2024 and onwards, the [RabbitMQ Core Team at Broadcom](https://github.com/rabbitmq/) only provides support to paying customers
-and [regularly contributing users](https://github.com/rabbitmq/rabbitmq-server/blob/main/COMMUNITY_SUPPORT.md).
+:::info[NEED ENTERPRISE SUPPORT?]
+VMware Tanzu RabbitMQ provides support for all RabbitMQ users on the latest 3 releases.
```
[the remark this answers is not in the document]
skwokz: Yup that makes sense. I'll update this so that it is clearer.

michaelklishin:
I have also ported the changes to the consulting partners page.

This leaves out the footer, some of the Release Information language (that was replaced), and the homepage which several members of our team specifically objected to changing in the proposed way.

I will update the announcement bar but like I said, it's not meant to be a permanent element and the user can hide it

src/pages/index.js
```diff
@@ -22,9 +20,25 @@ export default function Home() {
...
             </div>
-            <div class={styles.heroCta}>
-              <Link className="button button--primary" to="/tutorials">Getting Started</Link>
-              <Link className={styles.release_notes_link} to="https://github.com/rabbitmq/rabbitmq-server/releases/tag/v4.0.7">RabbitMQ 4.0.7</Link>
+          </div>
+        </div>
+        <div className={styles.why}>
+          <div className={styles.container}>
+            <Heading as="h1">Licensing</Heading>
+            <p>Since its original release in 2007, RabbitMQ is free and Open Source Software. In addition, Broadcom offers enterprise support and features.</p>
```
michaelklishin, line 28: So right on the home page, this is all we have to say about open source RabbitMQ? That it is open source software and that Broadcom offers commercial licenses?

Perhaps the home page should explain what RabbitMQ is, how it is often used, and then go into the licensing details?

skwokz: All the info about RMQ should still be on the homepage... unless I accidentally deleted it when I was trying to fix a merge conflict... The licensing information was moved up from the bottom of the page since the PMs were getting a lot of emails from people confused about the license (whether there is a free one or not, what the commercial offers etc). The "Why Rabbit MQ", "Examples of common use cases", and "What are users saying?" sections should all be intact. I'll double check when I resubmit this.

skwokz:
> The use of the temporary (collapsable) banner for such promos is fine but when a new version comes out, it will be replaced for at least a period of time. So it's a temporary banner the user can close and won't see until it changes.
 
> The changes to the Release Information page are acceptable as long as the release series table remains the central piece and not a yet another place to put some Tanzu RabbitMQ links (we already have some on that short page). I have ported over a few of these changes.

The goal here was to highlight commercial support for those who need it and also the make community support look more clear cut. The intent was NOT to make it look like there is no community support. As such the 3 paragraphs were combined into one and the "Important" banner was removed as it was seemed to no longer be timely. If there's key information to the community support that's missing or you have suggestions to make it clearer/look less diminished, I'm welcome the to feedback.

src/pages/index.js
michaelklishin: @skwokz those parts are still there but the most visible element is now this license statement plus a notice that there is a commercial license.

Those can be mentioned "in passing", for example, that open source RabbitMQ is MPLv2 licensed, with a commercial edition and longer term support available from Broadcom.

michaelklishin:
@skwokz I don't have a specific advise but our team's belief is that the homepage already displays most of what it should. Two more very commonly visited pages now have more (or more visible) Tanzu RabbitMQ and commercial support links.

The footer was updated to the extent we can: Docusaurus cannot handle anchors to JS pages unless those pages explicitly list that anchor, and we are no longer sure how that's done in the latest Docusaurus version.

Plus the announcement bar links to the commercial support Tanzu RabbitMQ page, and will stay that way for a good month now.

michaelklishin:
@skwokz another change that was rejected is the ✅ and ❌ indicators in the supported series table. Background colors is a reasonable way to highlight the (community) supported series in our opinion.

So what (if anything) to do with the home page is the only major outstanding item.

skwokz:
> @skwokz another change that was rejected is the ✅ and ❌ indicators in the supported series table. Background colors is a reasonable way to highlight the (community) supported series in our opinion.
> 
> So what (if anything) to do with the home page is the only major outstanding item.

Tl;dr rationale: In dark mode the white on yellow text fails WCAG standards. At minimum add an icon or find a bg color and text combination that passes WCAG standards. The changes I have go above and beyond, but if you choose to reject it, the key contrast issue should be addressed. 

Long explanation: Colored backgrounds behind text aren’t great for accessibility. The colors chosen in light mode is probably fine but in dark mode the text on the yellow background is likely illegible for people with common forms of color blindness. You can check contrast with a tool like https://webaim.org/resources/contrastchecker/. I have not checked the constrasts of the all the colors but that one in particular stood out. The icon adds an additional indicator for the state. For accessibility the minimum best practice is two indicators, in this case the date and color. The icon was to add a 3rd which is more than the minimum but improves the UX. Additionally, the goal was to simplify it with only two states.

michaelklishin:
@skwokz thank you for clarifying that part. I have no objections to moving away from colored backgrounds then.

Can we continue with smaller focused PRs, such as changing the visual aspects of the series tables to use emojis over the background color?

PR #2205, opened 2025-03-06
CONNECT_USING_PLAIN_HTTP Env Variable for the Messaging Topology Operator

okwilkins:
Nearly 2 years ago this issue was raised: https://github.com/rabbitmq/messaging-topology-operator/issues/621
With the following PR addressing it: https://github.com/rabbitmq/messaging-topology-operator/pull/622

This enabled the Messaging Topology Operator to communicate to RabbitMQ clusters via HTTP, even if the cluster had TLS enabled. The PR added in a new environment variable called `CONNECT_USING_PLAIN_HTTP`.

I don't believe that this was documented anywhere?

Furthermore, it also [addresses this limitation](https://www.rabbitmq.com/kubernetes/operator/tls-topology-operator#limitations):
> Messaging Topology Operator will not attempt to connect to the RabbitmqCluster over HTTP if HTTPS is available (even if the certificate is not trusted)

The need to remove this was also mentioned in the PR.

### DIFF
--- a/kubernetes/operator/install-topology-operator.md
+++ b/kubernetes/operator/install-topology-operator.md
@@ -264,6 +264,17 @@ The following table listes the Topology Operator environment variables that are 
       Default value is set to 2 seconds.
     </td>
   </tr>
+  <tr>
+    <td>
+      CONNECT_USING_PLAIN_HTTP
+    </td>
+    <td>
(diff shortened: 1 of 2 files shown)


michaelklishin (approved):
Thank you. This is a good catch.

PR #2206, opened 2025-03-07
Update Link to Working Delayed Messaging Implementation 

TravisNickels:
An existing link in the blog "scheduling messages with RabbitMQ" is broken, which could be confusing for users looking for a working implementation of delayed messaging with RabbitMQ.  This PR updates the link to point to the NServiceBus implementation, which is actively maintained and is widely used by our customers.  This should provide a more reliable reference for users choosing to implement delayed delivery themselves.


michaelklishin:
@TravisNickels I am really torn between retroactively changing a 10 year old blog post vs. linking to a resource that no longer exists.

Well, fair enough. NServiceBus is a good option to consider for the .NET community.

TravisNickels:
@michaelklishin, I totally understand not wanting to make a change to an older blog but appreciate the merge and your consideration.  Hopefully, this helps some readers in the future.  Thanks again!

PR #2207, opened 2025-03-08
Add note about poison messages with prefetch_count > 1

Arseniy-Popov:
This adds a warning that a single "poison message" may force other messages to be dropped or dead-lettered with `prefetch_count` > 1. This behavior follows straightforwardly from how prefetch counts and delivery counts work, but it is not immediately obvious to users and could lead to unexpected message loss.

### DIFF
--- a/versioned_docs/version-4.0/quorum-queues/index.md
+++ b/versioned_docs/version-4.0/quorum-queues/index.md
@@ -369,6 +369,12 @@ using an [optional queue argument](./queues#optional-arguments) at declaration t
 
 See [repeated requeues](#repeated-requeues) for more details.
 
+:::important
+
+Note that with `prefetch_count` > 1, if a consumer crashes due to a single "poison message" while other prefetched messages are not yet acknowledged, those other messages will also be redelivered and have their delivery counts incremented. This may force messages that could otherwise be processed normally (i.e. not "poison") to be dropped or dead-lettered.
+
+:::
+
 ### Configuring the Limit {#position-message-handling-configuring-limit}
 
 It is possible to set a delivery limit for a queue using a [policy](./parameters#policies) argument, `delivery-limit`.


michaelklishin:
Thank you but this is in no way quorum queue-specific. There is already an extensive documentation guide [on consumer acknowledgements](https://www.rabbitmq.com/docs/confirms#consumer-nacks-requeue).

michaelklishin:
> This may force messages that could otherwise be processed normally (i.e. not "poison") to be > dropped or dead-lettered.

This is incorrect, or at least not completely correct. Requeueing can result in almost immediate dead-lettering for the messages with a very short TTL but cannot result in messages being "dropped".

Arseniy-Popov:
Thank you for the prompt reply!

> Thank you but this is in no way quorum queue-specific.

Unless I'm mistaken, this behaviour is specific to quorum queues because only quorum queues track the delivery count and have a delivery limit. This note was placed in the section on poison message handling by quorum queues as this is the specific reason message loss may occur that is unique to quorum queues.

> Requeueing ... cannot result in messages being "dropped".

What if `x-delivery-count` exceeds `delivery-limit` and there isn't a dead-letter exchange configured? The docs themselves say that the message will be dropped:

> When a message has been redelivered more times than the limit the message will be dropped (removed) or dead-lettered (if a DLX is configured).

Source: [github.com/rabbitmq/rabbitmq-website/blob/main/versioned_docs/version-4.0/quorum-queues/index.md?plain=1#L350](http://github.com/rabbitmq/rabbitmq-website/blob/main/versioned_docs/version-4.0/quorum-queues/index.md?plain=1#L350)

michaelklishin:
@Arseniy-Popov if a delivery count exceeds delivery limit then the message is discarded. That's the whole point of the feature and [already stated right at the top of the section](https://www.rabbitmq.com/docs/quorum-queues#poison-message-handling).

This is not a typical ("normal") scenario and I don't see how it's relevant to the part I have called out, namely

> This may force messages that could otherwise be processed normally (i.e. not "poison") to be > dropped or dead-lettered

A message that did not exceed the redelivery limit will be re-queued.

What your PR tries to tell is that "this feature might throw away some data", which is the whole point of the feature, to eliminate a delivery-requeueing loop and let the quorum queue truncate its Raft log eventually.

Arseniy-Popov:
Consider the following scenario. There are three messages in a (quorum) queue. A consumer has a prefetch count of 3 and takes 3 messages and starts processing them concurrently. The processing involves a couple of seconds of I/O. Two of the messages are fine and can be processed successfully, at least in isolation. The third one immediately causes an OOM condition and a SIGKILL. Because the first two messages have not yet been processed, all three messages are requeued and have their delivery counts incremented. When the consumer is restarted it again receives the same three messages and again it's killed due to the third message before the two others are processed and again all three messages have their delivery counts incremented. This cycle continues until all three messages have their delivery counts reach the delivery limit and are dropped or dead-lettered.

So, we had just one "poison" message and yet it caused two "non-poison" messages to be dropped or dead-lettered. Had the prefetch count been at 1 the two "non-poison" messages would've been processed successfully. I believe this interplay between prefetch count and delivery limits deserves an explicit spelling out.

Arseniy-Popov:
@michaelklishin

michaelklishin:
@Arseniy-Popov an alternative (and very brief) wording was pushed for all website versions (3.13, 4.0, future 4.1) in https://github.com/rabbitmq/rabbitmq-website/commit/1845dd2e8a8ebcd47a8952a213c7b9c79fb85288.

PR #2213, opened 2025-03-17
Doc: Fix rules count for config syntax

glensc:
Commit 80d37d33c31ffe56d27e4a78f2e445dbc6b10684 added fourth rule.

### DIFF
--- a/docs/configure.md
+++ b/docs/configure.md
@@ -319,7 +319,7 @@ allows the RabbitMQ server and plugins to be configured.
 The file uses the [sysctl format](https://github.com/basho/cuttlefish/wiki/Cuttlefish-for-Application-Users),
 unlike `advanced.config` and the original `rabbitmq.config` (both use the Erlang terms format).
 
-The syntax can be briefly explained in 3 lines:
+The syntax can be briefly explained in 4 lines:
 
  * One setting uses one line
  * Lines are structured `Key = Value`


glensc:
And seems that not accurate either:
>  * Any line starting with a `#` character is a comment

by the description of the added fourth rule, It seems to trim everything (and including) the `#` character from lines.

I suppose `#` anywhere else than the first character wouldn't be a comment in the original description.

michaelklishin:
@glensc you [don't have to guess](https://github.com/Kyorai/cuttlefish/blob/main/src/conf_parse.peg), this is open source software.

glensc:
🚧 Never link to branches, share a permalink 🚧
- see https://news.ycombinator.com/item?id=8046710
  and https://help.github.com/articles/getting-permanent-links-to-files/
  and https://docs.gitlab.com/ee/user/shortcuts.html#project-files
- hit `y`
- and perhaps for https://github.com/josephfrazier/octopermalinker (GitHub)

NOTE: edit your post and fix it with a permalink!

michaelklishin:
@glensc Cuttlefish is a very slowly moving project, so linking to `main` is perfectly reasonable.

michaelklishin:
@glensc instead of telling the maintainers who to do or never to do, consider studying the README of the project you are contributing to so that we don't have to [finish your contributions for your because they were incomplete](https://github.com/rabbitmq/rabbitmq-website/commit/12e880dafa3246c2d8715e5ae3a97105c1be1af7).

PR #2214, opened 2025-03-18
[#922] controlling resources deletion

aawoznia26:
DeletionPolicy was added to `Federation`, `Queue`, `Shovel`, and `Vhost` resources. It controls whether resource in RabbitMQ cluster is removed when custom resource is deleted.
 The deletionPolicy can be set to:  
* `delete`: The resource will be removed from RabbitMQ cluster when the custom resource deleted (default)
* `retain`: The resource will remain in RabbitMQ cluster when the custom resource deleted

Corresponding [PR in Messaging Topology Operator repo](https://github.com/rabbitmq/messaging-topology-operator/pull/961)
Change is result of the [issue 922](https://github.com/rabbitmq/messaging-topology-operator/issues/922)


michaelklishin:
No, in that case I will revert the change. No worries, thank you for pointing that out.

PR #2217, opened 2025-03-20
Highlighted the use of the ssl_options.password option.

amusarra:
Reading the documentation and following the flow in order, you are led into error by not setting the password for the server certificate key right away, which is set following the example in the Using tls-gen's Basic Profile paragraph.

Usually it is rare to have a server certificate key without a password even for test/development environments.


lukebakken (approved):
Thank you.

PR #2220, opened 2025-03-26
Fix typo in exchanges.md

Valt25:
Fixed small typo. I am first time here, and I hope here is no a lot of rules to prepare PRs, and it can be easily merged due to simplicity of PR

### DIFF
--- a/docs/exchanges.md
+++ b/docs/exchanges.md
@@ -88,7 +88,7 @@ Topic exchanges use pattern matching of the [message's routing key](./publishers
 to the routing (binding) key pattern used at binding time.
 
 For the purpose of routing, the keys are separated into segments by `.`. Some segments are populated by specific values,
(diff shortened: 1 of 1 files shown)


michaelklishin:
@Valt25 thank you! This only updates the 4.1 version of the docs, I will apply the same change to 4.0.x (under `versioned_docs`) on your behalf.

michaelklishin:
Apparently the 4.0.x version did not have this typo. All good!

Valt25:
@michaelklishin Ok, now understand the versioning. And I would check and apply next changes on multiple versions.

However there is still same type in 4.0(https://github.com/rabbitmq/rabbitmq-website/blob/c1c357fff51dbc510f75ae25edcd246fd4703eac/docs/exchanges.md?plain=1#L91). And I see no type in "next"(4.1) version on docs https://www.rabbitmq.com/docs page(it's looking as changes have been applied). But 4.0 have this typo, should I prepare one more PR?

PR #2221, opened 2025-03-26
Fix typo in exchanges.md 4.0

Valt25:
I beleave there is still typo in 4.0 version.  Also I have checked version 3.13 and did not find such typo there


michaelklishin:
You are right. I've read the first word "segment" on that line and concluded that it was OK. Thank you for paying close attention.

PR #2244, opened 2025-04-16
Update using-topology-operator.md to have a real SHA-512 hash in the user example

jimlarnn:
The existing docs has a SHA-256 hash of "foobarbaz" in the example, even though the text says SHA-512


michaelklishin (approved):

michaelklishin:
Thank you.

PR #2255, opened 2025-04-30
Update download.md

pvladMQ:
Added a sub-chapter about minimum recommended system req for prod envs: 

## Minimum recommended system requirements for production (per node)


michaelklishin:
Closed because this belongs to a different guide, not because there's anything incorrect per se with the recommendation.

PR #2256, opened 2025-04-30
Add "minimum recommended system reqs"

pvladMQ:
Added minimum recommended system requirements subchapter to Clustering -> Cluster size. 

As per the recommendation in PR #2255, adding it here as it is a more appropriate location for this item.


michaelklishin:
@pvladMQ thank you!

PR #2261, opened 2025-05-08
Update protocol-listener health-check docs to support multiple protocols

the-mikedavis:
This is the companion change to https://github.com/rabbitmq/rabbitmq-server/pull/13871 which enables requesting multiple protocols in the health check.

Connects https://github.com/rabbitmq/rabbitmq-server/issues/13782


michaelklishin:
`4.1.1` is a week or two away but I think we can copy it to the 4.1.x version now merge  because that's a rarely used endpoint and it remains backwards compatible.

I'll do it.

PR #2270, opened 2025-05-20
Document RABBITMQ_SERVICE_PRIORITY environment variable for rabbitmq-service

inikulshin:
For https://github.com/rabbitmq/rabbitmq-server/pull/13919


michaelklishin:
Thank you but there doesn't seem to be much interest on the core team around making service priority configurable. I do not recall a single request for such a feature in my 15 years as a contributor.

PR #2271, opened 2025-05-20
fix:  resolved confusion with attaching log file

Abdur-Rahim-sheikh:
Currently if we run the receive_*.py file to dump logs in a log file, it does not reflect the message in the file immediately. Which creates confusion for a new practionner. added the warning and corrent command to show the reflection immediately in the file to resolve confusion.


michaelklishin:
I will manually apply this and credit you in the commit message. Thanks.

Abdur-Rahim-sheikh:
Thanks.
All I want is that other get a better experience with this tutorial.

PR #2274, opened 2025-05-24
docs(php): improve PHP documentation with the PHP7 capabilities

realihorrud:
Improved overall code style using PHP7 capabilities

### DIFF
--- a/tutorials/tutorial-five-php.md
+++ b/tutorials/tutorial-five-php.md
@@ -138,7 +138,7 @@ $channel = $connection->channel();
 
 $channel->exchange_declare('topic_logs', 'topic', false, false, false);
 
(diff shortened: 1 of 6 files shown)


michaelklishin:
@RudIhor thank you. Would you please take a look at the executable versions, too? https://github.com/rabbitmq/rabbitmq-tutorials

RudIhor:
@michaelklishin I have a little question: Can I update the whole style guide of code to match the latest PHP 8.4 features? I mean the following from the [RabbitMQ website](https://www.rabbitmq.com/tutorials/tutorial-one-php):
> This tutorial uses a client library that requires PHP 7.x or 8.x.

So if I can update the style guide of code to match the new PHP 8.4 features that would be much cooler than just updating it to PHP 7.0. Moreover it will show that RabbitMQ is on the pulse with PHP 8.4. Any thoughts?

PR #2276, opened 2025-05-27
Fix typo

mgrafl:


michaelklishin:
Thank you. 7c208ad7 correctly applies both changes to all editions, not just one for `main` and one for `4.1.x`.

PR #2277, opened 2025-05-28
Update tutorial-two-go-stream.md

streamdp:
fixed missed argument in the nested loop


michaelklishin (approved):

michaelklishin:
@streamdp thank you. May I ask you to also verify if the [executable version](https://github.com/rabbitmq/rabbitmq-tutorials) of these tutorials needs the same change?

PR #2279, opened 2025-05-29
Remove fail_if_no_peer_cert from LDAP documentation

ggustafsson:
Removed option `fail_if_no_peer_cert` from LDAP documentation because it is not a valid client option. 
https://www.erlang.org/doc/apps/ssl/ssl.html#t:client_option_cert/0

Removed option from all previous versioned docs and not just 3.13 since option has never worked.

Similar to:
- https://github.com/rabbitmq/rabbitmq-website/pull/1933
- https://github.com/rabbitmq/rabbitmq-website/commit/2ac6c0f96b0a3fa23a37b202c255f29e9bdd9f17

### DIFF
--- a/docs/ldap.md
+++ b/docs/ldap.md
@@ -363,11 +363,6 @@ auth_ldap.ssl_options.verify = verify_peer
 auth_ldap.ssl_options.verify = verify_none
 ```
 
-```ini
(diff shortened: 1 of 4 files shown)


michaelklishin (approved):

michaelklishin:
@ggustafsson thank you.

PR #2281, opened 2025-05-31
Update install-homebrew.md

specialorange:
added brew doctor and cleanup info for reinstall


michaelklishin:
Applied manually to all editions. Thank you.

PR #2291, opened 2025-06-25
Update connection-blocked.md

henryosei:
Updated documentation to avoid first person pronouns (eg. we, our).


michaelklishin:
Thank you.

PR #2315, opened 2025-07-18
Add clients compressions implementations

Gsantomaggio:
Documentation for clients' sub-entry batching compressions.

The compression happens client-side. Here we document the clients' implementations.

@michaelklishin @acogoluegnes if it is ok with you, I will copy it for the other versions. ( If necessary )


Gsantomaggio:
@michaelklishin changed `stream core` to `RabbitMQ core` and made the necessary fixes based on your suggestions.

michaelklishin (approved):

PR #2322, opened 2025-07-28
Adding Kourier, a Kotlin client

nathanfallet:
Hello!

I just added a Kotlin client under the Other JVM languages. Note that this client is compatible with Kotlin Multiplatform, so it also runs outside of JVM (maybe we should make a dedicated section for Kotlin?)

### DIFF
--- a/client-libraries/devtools.md
+++ b/client-libraries/devtools.md
@@ -200,6 +200,9 @@ Miscellaneous projects:
 
  * [March Hare, a JRuby RabbitMQ client](http://rubymarchhare.info)
 
+### Kotlin
+
+ * [Kourier, a Kotlin AMQP client](https://kourier.dev)
 
 ## C and C++ {#c-dev}
 
--- a/client-libraries/devtools.md
+++ b/client-libraries/devtools.md
@@ -202,7 +202,7 @@ Miscellaneous projects:
 
 ### Kotlin
 
- * [Kourier, a Kotlin AMQP client](https://kourier.dev)
+ * [Kourier, a Kotlin AMQP 0.9.1 client](https://kourier.dev)
 
 ## C and C++ {#c-dev}
 


michaelklishin (changes requested):
The domain name you are linking to does not exist.

It would be great if you could be more specific as to whether this is an AMQP 0-9-1 or AMQP 1.0 client. The two are completely different protocols.

michaelklishin:
@nathanfallet you should port the tutorials in the [executable tutorials repo](https://github.com/rabbitmq/rabbitmq-tutorials) first, then add them to the site and list them next to all other tutorials.

nathanfallet:
@michaelklishin What domain name do you have issue with? `kourier.dev` exists and works perfectly fine (I copy pasted from the changes and it works)
image

It's AMPQ 0.9.1, it's mentioned on the homepage as well as in the repository README. I can also change the link text to be more specific if this is what you mean?

nathanfallet:
Also, for the tutorial repo (I'll do it), there is an existing Kotlin folder, but it's using the Java client.
Do I update it using the pure Kotlin client (so it will be Kotlin for all available platforms) or do we make an additional folder so we keep Kotlin w/ Java Client (JVM only) AND Kotlin w/ Kotlin Client (all platforms)?
I would personally go for the first option since the public API is very close to the Java client one, but it supports Coroutines better and KMP so it might be the preferred one for Kotlin as of now 🤷

michaelklishin:
@nathanfallet sorry, `kourier.dev` indeed resolves on a "regular" machine with, say, `1.1.1.1` for DNS server but not when you are on the Broadcom VPN.

I'd say you should rename the `kotlin` directory to `kotlin-java-client` and use `kotlin` for this client because, well, now we actually use a Kotlin client.

nathanfallet:
Hey @michaelklishin it's me again!
Now the code tutorials are merged, anything I should know before proceeding with adding the tutorials on the public website?

nathanfallet:
Okay they are ready for review in #2406!

PR #2332, opened 2025-08-13
Tutorial correction: change "default queue" to "default exchange"

kirkland:


michaelklishin:
Thank you.

michaelklishin:
@kirkland thank you. I will push another edit that will update all the tutorials.

The intent of the phrase is not to suggest the default exchange but rather to use a server-named queue, even if the default exchange is used for replies by the request-reply server.

PR #2333, opened 2025-08-20
Document RABBITMQ_MAX_OPEN_FILES

mkuratczyk:
https://github.com/rabbitmq/rabbitmq-server/pull/14410


michaelklishin:
@mkuratczyk I will add a link to the existing relevant sections later today. We should mention that this is not the recommended way of managing the kernel limit but rather a workaround for certain environments.

Modern Linux distributions primarily use systemd unit files for controlling the limit, which is what we already document.

michaelklishin (approved):

PR #2335, opened 2025-08-21
Mention encrypted listeners in protocol-listener API reference

the-mikedavis:
The protocol-listener health check accepts protocol names with 's' for the encrypted listeners like `amqps` and `mqtts`. The health check distinguishes between encrypted and non-encrypted so to get a 200 when running AMQPS but not AMQP you need to query for `amqps`.

This change mentions the encrypted protocol names. I've also combined `amqp091` and `amqp10` as `amqp` which is what the health check translates these into.

Also see https://github.com/rabbitmq/rabbitmq-server/blob/c7f6cad331443df0eaad668c2f1ca62410259c3a/deps/rabbitmq_management/src/rabbit_mgmt_wm_health_check_protocol_listener.erl#L71-L135. The health check normalizes the protocol names so it can accept quite a large range

### DIFF
--- a/docs/http-api-reference.md
+++ b/docs/http-api-reference.md
@@ -1679,7 +1679,9 @@ Responds a 200 OK if if all given protocols have active listeners,
 otherwise responds with a 503 Service Unavailable. Multiple protocols
 may be provided by separating the names with commas.
 
-Valid protocol names are: `amqp091`, `amqp10`, `mqtt`, `stomp`, `web-mqtt`, `web-stomp`.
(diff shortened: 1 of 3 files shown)


the-mikedavis:
I applied the change to the 4.0 docs as well since even before https://github.com/rabbitmq/rabbitmq-server/pull/13871 (not backported to 4.0) the check accepted `amqps`, `mqtts`, etc - that stuff didn't change in https://github.com/rabbitmq/rabbitmq-server/pull/13871.

michaelklishin:
@the-mikedavis thank you!

PR #2344, opened 2025-08-28
fix unmatched CSS class names

Bowen7:
Some class names need to be renamed in the CSS file so that can work in dark mode

image


michaelklishin (approved):

michaelklishin:
Thank you.

PR #2352, opened 2025-09-04
Document the name of the credentials Secret generated for User objects

VannTen:
Knowing the Secret name before it is generated is required for some
workflows.

For instance, if you need to refer to the Secret to mount it in a Pod,
and you're not deploying your User object before writing your Pod spec
(common in gitops workflows), you need to have the Secret name
beforehand.

The Secret name is already fixed in code, but this is currently not
documented.

Update the documenation in accordance with the code.

Fixes rabbitmq/cluster-operator#1946

### DIFF
--- a/kubernetes/operator/using-topology-operator.md
+++ b/kubernetes/operator/using-topology-operator.md
@@ -310,6 +310,7 @@ The following manifest will create a user with generated username and password a
 accessed via a Kubernetes secret object:
 
 ```yaml
+---
 apiVersion: rabbitmq.com/v1beta1
 kind: User
 metadata:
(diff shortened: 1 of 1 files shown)


MirahImage (changes requested):

MirahImage:
There was previously a CLA in place under Pivotal and subsequently VMware. With the acquisition of VMware by Broadcom, the CLA signing bot that had previously enforced CLA signing was taken offline, but Broadcom have recently brought the CLA requirement back into force.

michaelklishin:
@VannTen this is not entirely new but there was a period of time where signing a CLA wasn't required, and now the requirement is back using different tooling. It's still a straightforward process (you sign using Box Sign) and the CLA is just over one page long, so it should not take too long to review.

We are OK with waiting. Thank you for contributing to RabbitMQ!

PR #2367, opened 2025-09-23
[RMQ-2157] Update columns and dates in support timelines table

Bowen7:
- Replace 'Next Release' with actual date
- In the 'Date of Release' column, replace patch release date with minor release date 
- Remove 'Patch' column
- Remove days from 'Date of Release' to only show month and year

We are providing 9 months of community support from minor release now, instead of until the next release. So I have added an ⁠`end_of_community_support` field to ⁠`4.1` in ⁠`releaseBranches`.


michaelklishin (changes requested):
@Bowen7 I suspect that `/release-information` also need to be updated.

In fact, that's arguably the main page that needs updating: the page has to do with open source RabbitMQ that actually has a community and a community support schedule :)

PR #2385, opened 2025-10-23
Document shovel rejections

dcorbacho:


michaelklishin:
@dcorbacho this behavior is not new so we should update all editions going back to at least 4.1.

See under `versioned_docs`.

michaelklishin:
Apparently the build failure was a genuine Cloudflare hiccup. Those happen once in a million years, too :)

PR #2393, opened 2025-11-14
update documentation for RabbitMQ - Prometheus integration

kenliao94:
Update instruction for running RabbitMQ <-> Prometheus integration with docker


michaelklishin:
@kenliao94 thank you.

PR #2422, opened 2026-01-04
Fix ae doc wording

torrca:
Remove redundant word in the alternate exchange documentation.


michaelklishin:
Good catch. Thank you for updating all the editions.

PR #2453, opened 2026-02-20
Fix: Operator policy example uses an invalid policy argument

decoyjoe:
The "patch" example was setting a `federation-upstream-set` argument, which is not one of the [valid operator policy arguments](https://www.rabbitmq.com/docs/policies#why-operator-policies-exist).

### DIFF
--- a/docs/policies.md
+++ b/docs/policies.md
@@ -1038,19 +1038,19 @@ rabbitmqadmin.exe operator_policies patch --name "cq.op-pol.1" --definition "{""
 The new `--definition` object will be merged into the existing policy definition.
 
 In the following example, an existing operator policy named `queues.op-pol.1` in the default virtual host (`/`)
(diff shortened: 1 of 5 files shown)


decoyjoe:
@michaelklishin it looks like there may still be a gap in the documentation. The [source](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_policies.erl#L50-L58) lists `queue-version` and `overflow` which don't appear in the [documentation](https://www.rabbitmq.com/docs/policies#why-operator-policies-exist). Additionally, the documentation lists `target-group-size` which isn't in the [source](https://github.com/rabbitmq/rabbitmq-server/blob/main/deps/rabbit/src/rabbit_policies.erl#L50-L58). If that's intentional then please disregard, just something I noticed.

michaelklishin:
@decoyjoe I will take a look, thanks.

PR #2499, opened 2026-03-31
Document the definitions.json file format

### LINKED ISSUE #1953: RabbitMQ definitions.json specification
### Is your feature request related to a problem? Please describe.

As a user, I may want to write custom definitions that get loaded when rabbitmq gets loaded, but there's no specification for the definitions.json file available in the documentation, or if there is I could not find it.

### Describe the solution you'd like

A clear description of the definitions.json file format/syntx, possible leveraging https://json-schema.org/ or similar definition syntax.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

lacatoire:
Fixes #1953

Adds a "Definitions File Format" section documenting the JSON structure used by definition export/import, with property tables and examples for all object types.

Applied to all editions (docs, 4.2, 4.1, 4.0, 3.13).

### DIFF
--- a/docs/definitions.md
+++ b/docs/definitions.md
@@ -50,6 +50,8 @@ Exported definitions use a JSON format. The top-level object contains the follow
 |-----|------|-------------|
 | `rabbit_version` | String | RabbitMQ version that produced the export |
 | `rabbitmq_version` | String | Same as `rabbit_version` (present for compatibility) |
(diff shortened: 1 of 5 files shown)


michaelklishin:
This is not a bad idea but we will likely take the machine readable format route, including validation commands for tools like `rabbitmqadmin` or `rabbitmqctl`.

Thank you for taking the time to contribute to RabbitMQ!

PR #2502, opened 2026-04-09
Document the verify_fun TLS option

### LINKED ISSUE #184: Document the verify_fun TLS option

lacatoire:
Document the Erlang/OTP `verify_fun` SSL option in the TLS guide for custom certificate verification during X.509 path validation.

Applied to all editions (docs/, 3.13, 4.0, 4.1, 4.2).

Closes #184


michaelklishin:
Said verification function needs to come from a plugin, and this PR does not cover that.

More importantly, custom verification functions are an increasingly rare phenomenon in the Erlang community. Usually tools have a reasonable default in place, or more opinionated plugins such as RabbitMQ's Trust Store one, based on this feature.

So I'm afraid we cannot accept this PR's "present scope".

PR #2509, opened 2026-04-20
Troubleshooting OAuth 2: add entry for "Not authorized" after login

### LINKED ISSUE #2055: Add new entry to the Oauth2 troubleshooting guide
### Is your feature request related to a problem? Please describe.

Sometimes a user who has successfully logged in to the management ui get an "Not authorized" error after they attempt to create a queue, or read from a queue. The user assumes that once it has successfully logged in and maybe has the administrator user tag, the user can do anything. 

### Describe the solution you'd like

The Oauth2 troubleshooting guide has already an entry with the title "No Authorized". However, this entry is for users who try to login for the first time and they get that error. However, there should be another "Not Authorized" entry for users who are already logged in but still get this error.

Ideally, the error "Not Authorized" should link directly to the troubleshooting guide. 

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

lacatoire:
Add a new troubleshooting entry for the "Not authorized" error a user can encounter after a successful management-UI login, when acting on a virtual host or a resource. This case is distinct from the existing "Not authorized" entry, which covers login-time errors.

The entry clarifies that management-UI tag scopes and per-resource permission scopes are independent, and summarises the scopes required by common actions (creating/deleting queues and exchanges, publishing, consuming, binding, managing policies).

Applied to `docs/` and to the 4.2, 4.1, 4.0 and 3.13 editions; 3.13's frame-size section is preserved.

Fixes #2055

### DIFF
--- a/docs/troubleshooting-oauth2.md
+++ b/docs/troubleshooting-oauth2.md
@@ -186,6 +186,72 @@ Follow these steps to find out which scopes or permissions are carried in the to
 11. Once you found the appropriate token's scope attribute, find within the attribute's value any of the scopes listed above. If [auth_oauth2.scope_prefix](./oauth2#scope-prefix) is used, it must be taken into account: the scopes will be named like  `myprefix_tag:administrator`. If [scope aliases](./oauth2-examples#using-scope-aliases) are used, find the scope alias that maps to one of the scopes listed above
 
 
+### Not authorized when performing an action {#not-authorized-after-login-error}
+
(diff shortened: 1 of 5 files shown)


michaelklishin:
After discussing this with the key maintainer of the OAuth 2 plugin we have concluded that some of the suggested updates are reasonable but we don't find the PR as a whole useful enough to adopt as it is.

Thank you for taking the time to contribute!

PR #2512, opened 2026-04-21
Explain the two types of scopes patterns

MarcialRosales:
Explain the two types of scopes patterns, which are wildcard and regexp.

Impl:https://github.com/rabbitmq/rabbitmq-server/pull/16196


michaelklishin:
https://github.com/rabbitmq/rabbitmq-server/pull/16196 is in => marking as Ready for review.

RichardJJG (approved):
Approving after minor edits

PR #2518, opened 2026-05-03
Add Repid library to devtools documentation

aleksul:


michaelklishin:
@aleksul this is a curated list. I haven't seen you announce this library anywhere, or anyone using it in RabbitMQ's own Discussions.

michaelklishin:
So the first release goes back to early 2023 and now it's at `2.0`. OK, fair enough.

aleksul:
@michaelklishin Thanks for merging! Sorry for the lack of context in the PR - 2.0 just launched and I hope it gets more traction soon. Also, I think it's worth adding that it's unique in that it provides its own asyncio-native implementation of AMQP 1.0, which differentiates it from most other Python libraries.

PR #2520, opened 2026-05-12
fixing a race condition in a kotlin tutorial

nathanfallet:
The original code wouldn't work in some cases the CPU is stressed (see the updated content, it explains itself). This is fixed in the [Kourier's doc](https://kourier.dev/tutorials/publisher-confirms.html) as well.


michaelklishin:
@nathanfallet thank you for taking the time.

Note that these are tutorials. They are meant to demonstrate one concept at a time (or a couple closely related ones) and are intentionally limited in scope.
None of the tutorials do sophisticated error handling, for example, for that reason.

Given that general idea for background, this change is difficult to justify.

If Kourier needs this for a straightforward code example to work, then Kourier should be fixed to handle such issues out of the box IMO.

PR #2526, opened 2026-05-19
Docs: quorum queue priorities in 4.3 — policy restriction and 32-level support

noahskelton:
Two related doc fixes for quorum queue priorities in 4.3.

The "What Cannot Be Set Using a Policy" section in the policies guide says the maximum number of priorities of *classic queues* cannot be set via a policy. As of 4.3, quorum queues also support priorities, and the same restriction applies to them — `max-priority` is not a recognised policy key for quorum queues either.

Reproduced against a 4.3.0 cluster with all feature flags enabled:

```
$ rabbitmqctl set_policy --apply-to quorum_queues test "^x$" '{"max-priority":5}'
Error: Validation failed
[{<<"max-priority">>,5}] are not recognised policy settings
```

Updated only the one bullet, in `docs/policies.md` and `versioned_docs/version-4.3/policies.md`.

The "Quorum Queues" subsection of `docs/priority.md` (and the 4.3 versioned copy) was also still describing pre-4.3 behaviour — only "two internal priorities (high/normal)" with priority 4 as the threshold, including an example table where priorities 1/2/8 dispatch as C, A, B. As of 4.3, quorum queues support 32 strict priority levels (0–31) via `x-max-priority`. Replaced the drifted text with the authoritative copy from `docs/quorum-queues.md`'s Priorities section, preserving the existing `### Quorum Queues` heading.

Older versioned copies (3.13, 4.0, 4.1, 4.2) are left untouched since strict quorum queue priorities are a 4.3 feature.

### DIFF
--- a/docs/policies.md
+++ b/docs/policies.md
@@ -131,7 +131,7 @@ Some optional arguments cannot be configured using a policy. They control variou
 that cannot be changed at runtime. Two key examples are:
 
  * The type of a queue
- * The [maximum number of priorities](./priority) of classic queues
+ * The [maximum number of priorities](./priority) of classic and quorum queues
 
 Those values intentionally cannot be configured by policies: their values are fixed at queue declaration time.
 
--- a/docs/priority.md
+++ b/docs/priority.md
@@ -131,22 +131,35 @@ priority.
 
 ### Quorum Queues
 
-Quorum queues internally only support two priorities: high and normal. Messages without
-a priority set or having priorities in the [0, 4] range will be considered to be of normal priority.
+:::important
 
-Messages with a priority higher than 4 will be considered to be of high priority.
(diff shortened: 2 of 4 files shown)


noahskelton:
Pushed a second commit (b1a4257) updating the "Quorum Queues" subsection of `docs/priority.md` and `versioned_docs/version-4.3/priority.md`. The text was still describing pre-4.3 behaviour (two internal priorities, threshold at 4, C/A/B example). Replaced it with the authoritative copy from `docs/quorum-queues.md`'s Priorities section, which documents the 32 strict priority levels (0–31) introduced in 4.3. PR title/body updated to cover both fixes.

michaelklishin:
Thank you.

PR #2541, opened 2026-06-11
Doc update: strongly advise against manual CLI cluster formation

pvladMQ:
Updated the cluster formation documentation to explicitly recommend automated peer discovery for production environments. The previous wording ("Consider using...") was too soft and led users to rely on manual `rabbitmqctl join_cluster` commands in production. 

Added a warning that manual CLI joins are prone to human error. Discussed with Mirah.


MirahImage (approved):

michaelklishin:
This only covers one edition (4.3). I will extend to the rest all the way to 3.13.

PR #2552, opened 2026-07-02
Document Topology Operator new feature: User password rotation

Zerpet:
This PR documents the feature introduced in rabbitmq/messaging-topology-operator#1205

This feature allows to rotate RabbitMQ User passwords by updating a `Secret` in Kubernetes.
This enables use cases where the `Secret` with user credentials is managed by an external
operator e.g. External Secrets Operator (ESO) or Vault.

This feature also introduces a new requirement on importing secrets: they must be labelled
with a specific label, otherwise they'll be rejected with a clear error message. This is
important because the operator keeps an index in memory of all resources it watches. In order
to avoid caching *all* secrets in a cluster, the operator set a label filter, so that only
secrets with a specific label are kept in the index (i.e. informer's cache).

### DIFF
--- a/kubernetes/operator/using-topology-operator.md
+++ b/kubernetes/operator/using-topology-operator.md
@@ -11,6 +11,7 @@ If RabbitMQ Messaging Topology Operator is not installed, see the [quickstart in
 
 This information includes the following sections:
 
+* [Secret Label Requirement](#secret-label-requirement)
 * [Cluster Operator Requirements](#requirements)
 * [Scope across multiple namespaces](#namespace-scope)
 * [Non Operator managed RabbitMQ](#non-operator)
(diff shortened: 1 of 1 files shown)


kubernetes/operator/using-topology-operator.md
```diff
@@ -355,14 +385,25 @@ As a workaround, add a label or annotation to `users.rabbitmq.com` object to tri
 #### Users with provided username and password {#provided-sec-users}
 
 The Operator also supports creating RabbitMQ users with provided credentials. When creating a user with provided username and password, create a kubernetes
-secret object containing keys `username` and `password` in its Data field. The Operator will then generate a new secret with the name of the User object suffixed with '-user-credentials'.
-The Operator does not monitor either the provided secret object or the generated secret object, and updating either secret object won't update the credentials.
-As a workaround, edit the generated secret object, then add a label or annotation to `users.rabbitmq.com` object to trigger the Operator to reconcile.
+secret object containing keys `username` and `password` in its Data field.
+
+**Important:** All secrets referenced via `importCredentialsSecret` must be labeled with `rabbitmq.com/topology-operator: "true"`. The Operator's Secret cache is restricted to labeled secrets, and the admission webhook will reject a User that references an unlabeled or missing secret.
```
michaelklishin: This should be an [admonition](https://docusaurus.io/docs/markdown-features/admonitions), no?

michaelklishin (changes requested):

kubernetes/operator/using-topology-operator.md
Zerpet: Indeed, it should. Pushed a change!

michaelklishin (approved):

PR #2559, opened 2026-07-08
docs: Improve the instruction to be more clearer

ammarsaf:
Improve the instruction to be more clearer


michaelklishin:
Thank you but I don't see why "the flag" would need explaining in this case. The sentence immediately before mentions the option name.

PR #2571, opened 2026-07-16
Fix non-deterministic sorting of security advisories

ansd:
The `fetch-security-advisories.sh` script previously sorted advisories only by their `published_at` date, which caused daily diffs when multiple advisories were published on the same day and the GitHub API returned them in a different order. This change adds the `repo` and `ghsa_id` as a secondary sorting keys to ensure a stable deterministic output order.


michaelklishin (approved):
Lexicographical sorting for a given date sounds very reasonable.

PR #2582, opened 2026-07-22
Fix dependabot updates

Zerpet:
The overrides entry pinned an independent version range from the
devDependencies entry. When Dependabot bumps the direct dependency
across a major version, npm requires overrides targeting a direct
dependency to match it exactly, causing an EOVERRIDE failure.
Referencing it via $webpack-dev-server keeps it in sync automatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "esbuild": "^0.25.0",
     "minimatch": "~3.1.3",
     "http-proxy-middleware": "^3.0.6",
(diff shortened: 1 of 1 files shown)


michaelklishin:
@Zerpet thank you. There was another dependency where the same fix would be appropriate.

I have a couple more ideas in mind, namely dependency grouping, but they can go into a follow-up PR.

michaelklishin (approved):

PR #2585, opened 2026-07-23
Document log term depth limit settings

lukebakken:
## Summary

Documents the log term depth limit settings, which cap the depth to which nested Erlang terms are printed in log messages so that deeper sub-terms are truncated to `...`.

A new "Term Depth Limit" section in the advanced log format documentation covers:

- the per-output `log.file.formatter.depth`, `log.console.formatter.depth`, `log.exchange.formatter.depth` and `log.syslog.formatter.depth` settings, their valid values (`unlimited` or a positive integer) and floor behavior (explicit values below 5 are raised to 5)
- the older, global `log.error_logger_format_depth` setting these fall back to, including its mapping to the Erlang `kernel` variable, its floor (integers below 10 are raised to 10), and how the two interact (per-output settings take precedence)

Neither setting was previously documented.

## Related

Documents the feature added in rabbitmq/rabbitmq-server#16992.

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -721,6 +721,57 @@ The following values are valid:
 * `lc4`: four characters, lower case, e.g. `dbug` or `warn`
 * `uc4`: four characters, upper case, e.g. `DBUG` or `WARN`
 
+### Term Depth Limit {#depth}
+
+Some log messages contain deeply nested Erlang terms, such as large data
+structures embedded in crash reports. The depth to which such terms are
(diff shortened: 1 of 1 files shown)


michaelklishin (approved):

michaelklishin:
I had to revert: this was a part of https://github.com/rabbitmq/rabbitmq-server/pull/16992 and not some docs for prior changes, which is how I've interpreted the change at first.

PR #2589, opened 2026-07-25
Update exchange name in tutorial 4 and Java stream source file

ngocnhan-tran1996:
I think the exchange name should be `X` instead.

image

Reference
https://www.rabbitmq.com/tutorials/tutorial-four-java#direct-exchange


michaelklishin:
Thank you.

PR #2637, opened 2026-08-31
Include TLS settings in example rabbitmqadmin configuration

tyilo:
The example is taken from https://github.com/rabbitmq/rabbitmqadmin-ng#configuration-file-format-moved-to-toml


michaelklishin:
Thank you.

PR #2649, opened 2026-09-15
Add rabbitmq_auth_backend_aoptoken to community plugins

martinx:
This adds a new **Authentication** section listing `rabbitmq_auth_backend_aoptoken`.

### What the plugin does
It accepts a pre-issued JWT bearer token (`token:<jwt>`, carrying a `sub` claim — the credential style issued by Apache Pulsar and AoP-based brokers) in the AMQP password field, **alongside** internal username/password authentication on the **same account**. This lets clients migrating from such deployments onto native RabbitMQ keep authenticating without any client-side change.

It is authentication-only: put it in a chain and delegate authorization to the internal backend, so permissions keep coming from RabbitMQ's own database.

```ini
auth_backends.1 = internal
auth_backends.2.authn = rabbit_auth_backend_aoptoken
auth_backends.2.authz = internal
```

### Details
- Algorithms: `HS256/384/512` with a shared secret, `RS256/384/512` and `ES256/384/512` with a public key — matching Pulsar's `tokenSecretKey` and `tokenPublicKey` modes. `alg: none` and unknown algorithms are rejected.
- `exp` and `nbf` are enforced when present (optional leeway); an expected `aud` can be required.
- Keys are configurable and never embedded in the plugin.
- Repository: https://github.com/martinx/rabbitmq-auth-backend-aoptoken
- Documentation: https://aoptoken.bitey.ai
- License: MIT
- Tests: 28 EUnit cases covering every algorithm and rejection path (tampering, algorithm confusion, expiry, audience, missing configuration). CI runs the suite and packages the plugin against RabbitMQ **3.11.28, 3.12.14, 3.13.7, 4.0.5 and 4.1.1**.
- Prebuilt `.ez` assets are attached to each GitHub release (OTP 26 and 27).

I placed it under a new "Authentication" section. Happy to move it (e.g. next to the existing `rabbitmq_auth_backend_ip_range` entry) or adjust the wording if you prefer.


michaelklishin:
Thanks.

We will have to try and review the plugin, which will take longer than editing the wording in a short new HTML table. Sorry.

